PR #12, opened 2016-05-20
gitignore structured log files

AArnott:
Your .gitattributes policy of not normalizing line endings has already led to inconsistent line endings in this file. You had some LF and some CRLF. So I conformed to LF and fixed the line you added with CRLF to be consistent.


KirillOsenkov:
Thanks for fixing this. I believe it shouldn't be Git's job to mess with content. Instead, the Editor should help with that. I'm going to make sure the editor I use will assist with that :)

AArnott:
> I believe it shouldn't be Git's job to mess with content.

Be sure to also break git merge so that it never touches your content. It should leave all conflicts unresolved so you can fix them yourself with your editor. ;-p

KirillOsenkov:
I merge through VS and let VS handle all that ;)

AArnott:
VS uses git internally. Today it's libgit2, but tomorrow, it's git.exe. :)

PR #13, opened 2016-05-20
Use Nerdbank.GitVersioning

AArnott:
This replaces manual bumping of versions with automatic incrementing of the 3rd version component.


AArnott:
dang. you're quick. :)

PR #34, opened 2016-07-21
Trigger BuildCommand when pressing enter in Build Arguments

### LINKED ISSUE #33: Pressing ENTER on build arguments page should start the build

rahulpnath:
This fixes #33. 

- Added a KeyBinding to the textbox to trigger the BuildCommand when pressing enter key in the arguments textbox. 
- Set the UpdateSourceTrigger to PropertyChanged so that the binding is updated immediately as earlier it was getting triggered automatically on loosing focus , which will no longer happen if pressing Enter key.


KirillOsenkov:
Oh! How very nice. Thank you!

PR #70, opened 2017-05-17
Optimize message loading

pakrym:
`AddMessage` does a lot of calls to `GetOrCreateNodeWithName` which allocates enumerator, lambda closure, enumerates all children and tries to cast them.  It's unable to open 30mb binlog file on my machine. This change adds a temporary (until `Seal`) cache of child nodes.

### DIFF
--- a/src/StructuredLogger/ObjectModel/TreeNode.cs
+++ b/src/StructuredLogger/ObjectModel/TreeNode.cs
@@ -47,6 +47,8 @@ namespace Microsoft.Build.Logging.StructuredLogger
             }
         }
 
+        private Dictionary<ChildrenCacheKey, object> childrenCache;
+
         private IList<object> children;
         public bool HasChildren => children != null && children.Count > 0;
 
@@ -69,6 +71,7 @@ namespace Microsoft.Build.Logging.StructuredLogger
             {
                 children = children.ToArray();
             }
+            childrenCache = null;
         }
 
         public void AddChildAtBeginning(object child)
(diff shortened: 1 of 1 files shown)


pakrym:
In reality "before" never finished, so it's much faster then this comparison shows.
image

src/StructuredLogger/ObjectModel/TreeNode.cs
```diff
@@ -47,6 +47,8 @@ public bool IsExpanded
             }
         }
 
+        private Dictionary<ChildrenCacheKey, object> childrenCache;
```
KirillOsenkov: My only concern is that every object in the tree will become 8 bytes larger (the viewer is a 64-bit app) due to this new field. I've worked very hard to fine tune the object layout and minimize fields on base classes such as TreeNode.

How about instead of storing the childrenCache in a separate field you make a class ChildrenList : List<object> which will also store the dictionary and the caching logic. Then you can just switch to using your custom collection as storage for Children, and in the Seal() method when you ToArray() that, it just goes away and the unnecessary dictionary will get collected.

What do you think about that?

KirillOsenkov:
Overall I love this change and thanks a lot for profiling. See if you think you can avoid adding an extra field on the tree class. I'd be super curious to see the final memory layout with your current change and after you get rid of the field, I'm sure the'll be a different in memory consumption (after GC).

src/StructuredLogger/ObjectModel/TreeNode.cs
pakrym: GetOrCreateNodeWithName needs access to cache, so you can not just store IList and access cache at the same time.

pakrym: I can add and abstraction and two implementations - one array based and one list based.

KirillOsenkov: You can just change the type IList to be specific type ChildrenList?

KirillOsenkov:
Thanks @pakrym I've used your PR as an inspiration to introduce ChildrenList, hopefully it will improve performance for you.

PR #111, opened 2017-08-18
Fix typo in README.md

craigb:
reference the StructuredLogger.dll, there is no BinaryLogger.dll

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ Alternatively (useful for older versions of MSBuild) you can attach the logger t
 https://www.nuget.org/packages/Microsoft.Build.Logging.StructuredLogger
 
 ```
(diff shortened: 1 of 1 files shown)


KirillOsenkov:
Nope, this is correct. There are two loggers, StructuredLogger and BinaryLogger, both in StructuredLogger.dll (sorry I know it's confusing). Here's a write-up on the difference: 
https://github.com/KirillOsenkov/MSBuildStructuredLog#differences-between-structuredlogger-and-the-new-binarylogger-that-is-shipped-with-msbuild-as-of-153

We recommend using the BinaryLogger.

KirillOsenkov:
Hey @craigb I wanted to apologize for the confusion! You were correct and I was wrong! There is indeed only one .dll, StructuredLogger.dll, so thanks for submitting this.

I don't know what I was thinking, maybe I was on the phone or distracted. You were correct of course.

PR #130, opened 2017-10-21
Fix typo in README.md

craigb:
Reference the StructuredLogger.dll, there is no BinaryLogger.dll


KirillOsenkov:
Thanks! I can't believe I still had that typo in another place, even though we already corrected it once. I'm getting old.

PR #190, opened 2018-08-18
Split logic

decriptor:
@KirillOsenkov This is a fairly quick initial pass through moving out non-UI specific code out of StructuredLogViewer. I also went through and switched over to GuiLabs.Language.Xml along with bumping other nuget packages. I didn't bump the MS Build nugets. I'm not sure if you wanted to support a newer .net framework, etc.

I wanted to get an early review of all of this before doing too much more. The long term goal is to add an XWT frontend.


KirillOsenkov:
We can either downgrade Squirrel back to what master is using, or update the .nuspec file to use the prerequisites required by Squirrel.dll 1.8.0.0.

PR #197, opened 2018-09-11
Massive API breaking change to remove conflicts with MSBuild.

KirillOsenkov:
Previously we used to keep a fork of the same public types and namespaces present in MSBuild (such as BinaryLogger, BinaryLogReplayEventSource, etc). It was easier to sync with the MSBuild codebase.

However this presents a problem - if someone is referencing both MSBuild and StructuredLogger.dll then they have a conflict - types with the same namespace and name are present in both places, so it fails the build.

I've decided to move public types which are mirrored from MSBuild into a different namespace, and additionally rename BinaryLogReplayEventSource -> BinLogReader.

Also switching the version from 1.2 to 2.0 as this is a major breaking change.

### DIFF
--- a/src/StructuredLogger.Tests/Roundtrip.cs
+++ b/src/StructuredLogger.Tests/Roundtrip.cs
@@ -61,7 +61,7 @@ namespace StructuredLogger.Tests
         //[Fact]
         public void ReadBinaryLogRecords()
         {
-            var reader = new BinaryLogReplayEventSource();
+            var reader = new BinLogReader();
             var records = reader.ReadRecords(@"C:\temp\msbuild.binlog");
             foreach (var record in records)
             {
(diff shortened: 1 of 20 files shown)


KirillOsenkov:
Code review please?

@AndyGerlicher @rainersigwald @daveaglick @pascalberger @rodrmoya

pascalberger:
@KirillOsenkov Don't know really the source code, but changes make sense from what I can see. If you wan't I can also build and test it with my source code.

pascalberger:
@KirillOsenkov I tested the changes in my code base and everything seems to work fine

KirillOsenkov:
Published: https://www.nuget.org/packages/MSBuild.StructuredLogger/2.0.4

pascalberger:
@KirillOsenkov Thank you very much for the fast fix. Really appreciated 👍

PR #225, opened 2018-12-02
Detect double writes from compilation

### LINKED ISSUE #217: Feature request: Double writes during compilation
The double write logger mostly operates by looking at the invocations of `CopyTask`. That's very effective but misses out on tasks like `Csc` and `Vbc`. 

More than once I've gotten bitten because we messed up our build scripts and suddenly started writing build output (DLL, XML, PDB) to the same place during a multi-targeted build. Wanted to add logic to detect this. 

I'm unsure if this is the right place or not. `CopyTask` is a built-in task hence we know what `CopyTask` is and what semantics it has. The `Csc` task though is not built-in, although it's about as close as can be. Hence unsure if this is too specific for this repository.

Happy to do the work here BTW. If this isn't the right place I'm just going to add it to Roslyn directly :smile:

jaredpar:
This adds support for detecting double writes that occur during a compilation from the C#, VB and F# compilers. 

For single targeted projects this isn't much of a problem. After all if you have two different projects compiling the same assembly to the same directory then you will likely notice this real fast. 

For multi targeted projects though this is a real problem. It's very easy to leave off a `$(TargetFramework)` sub-directory in a path and say write the same XML documentation file for every target framework. This change helps detect mistakes like that which otherwise tend to go unnoticed.

Fixes #217

### DIFF
--- a/src/StructuredLogger/Analyzers/DoubleWritesAnalyzer.cs
+++ b/src/StructuredLogger/Analyzers/DoubleWritesAnalyzer.cs
@@ -42,9 +42,13 @@ namespace Microsoft.Build.Logging.StructuredLogger
             {
                 AnalyzeCopyTask(copyTask);
             }
-            else if (task is CscTask cscTask)
+            else if (task is CscTask cscTask && cscTask.CompilationWrites.HasValue)
             {
-                AnalyzeCscTask(cscTask);
+                AnalyzeCompilationWrites(cscTask.CompilationWrites.Value);
+            }
+            else if (task is VbcTask vbcTask && vbcTask.CompilationWrites.HasValue)
+            {
+                AnalyzeCompilationWrites(vbcTask.CompilationWrites.Value);
             }
         }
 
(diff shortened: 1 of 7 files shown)


src/StructuredLogger/Analyzers/DoubleWritesAnalyzer.cs
```diff
@@ -36,7 +36,27 @@ public void AppendDoubleWritesFolder(Build build)
             }
         }
 
-        public void AnalyzeFileCopies(CopyTask copyTask)
```
jaredpar: This does change an API that was previously public. I'm unsure how critical this is or isn't for this component. Can add it back if needed.

KirillOsenkov:
Thanks! I've extracted a base class ManagedCompilerTask too.

Uploaded the NuGet package to https://www.nuget.org/packages/MSBuild.StructuredLogger/2.0.61. Note that the NuGet package ID changed.

PR #233, opened 2019-01-17
Added serialization for Import and NoImport types

xen2:
Had to do this to properly save and reopen my binlog (it would otherwise crash on opening on `Activator.CreateInstance` because `Import` and `NoImport` had no ctor).


KirillOsenkov:
Nice, thanks!

KirillOsenkov:
This fix is now deployed in 2.0.68.

PR #236, opened 2019-01-30
Make search results stand out

ltrzesniewski:
I wanted the search results to be easier to navigate by making them stand out in the main tree.

Here's what this PR does (in this example, I searched for *"compile"*):

image

As you can see, there are two kinds of markers:

- image means the node is a search result
- image means the node *contains* a search result

Feel free to change the colors if you can find something better :)

As for the implementation, I noticed [a comment in `BaseNode`](https://github.com/KirillOsenkov/MSBuildStructuredLog/blob/fc260682c4a43123faa516c9c8b38c20a69bfca4/src/StructuredLogger/ObjectModel/BaseNode.cs#L5-L10) which says that the node size should be kept at a minimum, so I grouped the necessary data under a flags enum (1 byte). I also removed the underlying fields from `IsVisible`/`IsExpanded` so this change should make the tree use *less* memory than before while adding a new feature. If you accept this PR, I'll also add `IsLowRelevance` to the flags later to save another bool.

### DIFF
--- a/src/StructuredLogViewer.Core/Search.cs
+++ b/src/StructuredLogViewer.Core/Search.cs
@@ -20,6 +20,8 @@ namespace StructuredLogViewer
 
         public IEnumerable<SearchResult> FindNodes(string query)
         {
+            ClearSearchResults(build);
+
             var matcher = new NodeQueryMatcher(query, build.StringTable.Instances);
 
             resultSet = new List<SearchResult>();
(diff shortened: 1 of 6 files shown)


KirillOsenkov:
Excellent! Thank you very much! Nice clean change and works great! I appreciate that you've thought about the field packing. When you have one bool, you can add three more for free on 32-bit (due to alignment), but I agree Flags is much nicer.

PR #239, opened 2019-03-13
Support Ctrl+MouseWheel zooming UI

drewnoakes:
Useful when screen sharing or presenting.


colombod:
Maybe a way to reset the zoom as well? VS Code uses CTRL+0 to restore scale

KirillOsenkov:
I've implemented Ctrl+0 to reset the zoom level with this commit: https://github.com/KirillOsenkov/MSBuildStructuredLog/commit/5853e20b0212bdbe7642dc7cd1a7a5ef0f31ab4a

Thanks for the contribution!

Issue #87, opened 2017-07-12
Syntax highlighting for *.sln files


VladimirReshetnikov:
It would be great to have some syntax highlighting for *.sln files.
Also expanding-collapsing regions.

PR #247, opened 2019-06-06
Add syntax highlighting for .sln files

drewnoakes:
Fixes half of #87 (highlighting, but no folding).

I was curious to see how syntax highlighting worked in the Avalonia editor. Turns out to be pretty simple. I chose colours to match project files. Could probably do something fancier with the highlighting file, but this seems to work well enough.

image

### DIFF
--- a/src/StructuredLogViewer/Controls/TextViewerControl.xaml.cs
+++ b/src/StructuredLogViewer/Controls/TextViewerControl.xaml.cs
@@ -2,19 +2,25 @@
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
+using System.Reflection;
+using System.Text.RegularExpressions;
 using System.Windows;
 using System.Windows.Controls;
 using System.Windows.Input;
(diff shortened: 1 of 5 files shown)


src/StructuredLogViewer/Controls/TextViewerControl.xaml.cs
```diff
@@ -72,6 +78,19 @@ public void SetText(string text)
...
             }
 
+            if (solutionFileRegex.IsMatch(text))
+            {
+                IsXml = false;
+
+                using (var stream = Assembly.GetExecutingAssembly().GetManifestResourceStream("StructuredLogViewer.Resources.SolutionFile.xshd"))
+                using (var reader = XmlReader.Create(stream))
+                {
+                    textEditor.SyntaxHighlighting = HighlightingLoader.Load(reader, HighlightingManager.Instance);
```
KirillOsenkov: we could cache the highlighting for the next time, but no biggie

PR #249, opened 2019-07-05
Shift and mouse wheel scrolls UI horizontally

drewnoakes:
This is a feature of many applications and is very handy once you get used to it.


KirillOsenkov:
Thanks for your contribution! Published an update, 2.0.103.

PR #251, opened 2019-07-22
Let Windows prompt the user for an application to Open a file with

davidwengier:
Currently when the Open button is used on a `.targets` or `.props` file, when there is no application mapped to open those files, an exception is thrown. This PR tells Windows to show the standard dialog for unmapped files, like so:

image


drewnoakes:
Nice.

PR #263, opened 2019-10-12
Some updates to the Avalonia viewer

ltrzesniewski:
I wanted to take a look at AvaloniaUI, and this project offered a nice opportunity :)

Here's what I did:

 - Updated Avalonia to the latest v0.8.3
 - Ported my previous change #236 to the Avalonia viewer
 - Added support for Import/NoImport log items
 - Fixed the recent links on the start page
 - Fixed the project icons and item gradients

It's still not as pretty as the WPF viewer, but it's a bit better than before:

image

I hope I didn't break anything, since I'm not familiar with Avalonia, but everything I tested seems to work.


KirillOsenkov:
Thanks!

PR #264, opened 2019-10-13
Improve search in the Avalonia viewer

ltrzesniewski:
Here's another round of changes, which aims to significantly improve the search feature in the Avalonia viewer:

- Added the *"show all results"* button
- Added the *"elapsed time"* message
- Search results are now collapsed (Avalonia's `TreeView` does not support virtualization)
- Fixed the node icons in the search results (they were always set to the *Property* icon before)
- Navigate when a search result is selected (for the *"search log"* and *"find in files"* tabs)
- Fixed keyboard shortcuts (so Ctrl+F now works)

This also includes some minor changes in the WPF viewer, for consistency and cleanup:

 - Introduced a `NodeIcon` style to reduce duplication
 - Fixed text alignment on project nodes (whose icons are a  bit taller)

### DIFF
--- a/src/StructuredLogViewer.Avalonia/App.xaml
+++ b/src/StructuredLogViewer.Avalonia/App.xaml
@@ -41,6 +41,21 @@
                 </ControlTemplate>
             </Setter>
         </Style>
+        
+        <Style Selector="Rectangle.nodeIcon">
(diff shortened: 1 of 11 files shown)


KirillOsenkov:
:+1:

PR #266, opened 2019-10-20
Improve the log tree in the Avalonia viewer

ltrzesniewski:
Here are some more fixes and tweaks for the Avalonia viewer.

This mainly improves the log tree to make it look more similar to the WPF version:

 - Added node colors
 - Added the low relevance flag
 - Removed the padding to increase the information density

Here's what it looks like now:

image

Other changes:

 - Target .NET Core 3.0
 - Replaced the custom `TabItemsControl` with the standard `TabControl`
 - Fixed the tab headers in the source view (but the tabs still don't wrap correctly - that looks like an Avalonia bug)
 - Added the `ParentAllTargetsUnderProject` setting


KirillOsenkov:
Thank you so much!

PR #270, opened 2019-11-02
Presents targets in Files tree

cezarypiatek:
I've added targets on the FIles's tree.  This makes the target exploration very easy.

image


KirillOsenkov:
This is neat! Thanks!

PR #271, opened 2019-11-04
Add location of MsBuild from VS2019

cezarypiatek:
I've removed some of the hardcoded paths and I've used `Microsoft.Build.Locator` for discovering MSBuild instances. Now I'm able to select MSBuild from VS2019 from the suggestion list. No sure, but maybe this  Not sure, but maybe the whole logic of `GetMSBuildLocations` should be implemented using `Microsoft.Build.Locator` - what I mean is to completely get rid of hardcoded paths (at least for Visual Studio).

image


KirillOsenkov:
Thanks!

PR #293, opened 2020-02-10
Update to net472 and newer msbuild

cdmihai:
This allows me to reference the netstandard2 MSAGL libraries.


KirillOsenkov:
I just realized that by moving from MSBuild 14 references for desktop we lose the ability to use StructuredLogger.dll as a logger in MSBuild 14. I think it's time to drop that support anyway. They can find and use an older .dll if they need to.

PR #302, opened 2020-02-27
Sort runtime graph children and fix common properties

cdmihai:
Fixes:
- Sort children project nodes by `Project.StartTime`. This aids in understanding the flow and order of execution.
- Common global property node is computed right this time. It is also placed on top of root nodes, instead of sometimes being awkwardly placed in the middle of the graph.

Before:
image

After:
image


cdmihai:
Woops, I forgot to add back entry targets for root nodes, don't merge until that happens :)

cdmihai:
> Woops, I forgot to add back entry targets for root nodes, don't merge until that happens :)

Entry nodes added back, ready for review and merge.

PR #303, opened 2020-02-28
Fix dictionary key collision in RuntimeGraphNode

cdmihai:
Sometimes msbuild give the same start time for different Projects, which caused a key already exists exception. Fixed by using a nearest non conflicting start time.

### DIFF
--- a/src/StructuredLogViewer.Core/ProjectGraph/RuntimeGraph.cs
+++ b/src/StructuredLogViewer.Core/ProjectGraph/RuntimeGraph.cs
@@ -13,7 +13,7 @@ namespace StructuredLogViewer.Core.ProjectGraph
 {
     public class RuntimeGraph
     {
-        [DebuggerDisplay("{DebuggerDisplayString()}")]
+        [DebuggerDisplay("{ToString()}")]
         public class RuntimeGraphNode
         {
             private SortedList<DateTime, RuntimeGraphNode>? sortedChildren;
@@ -43,11 +43,11 @@ namespace StructuredLogViewer.Core.ProjectGraph
                 Project = p;
             }
 
-            public string DebuggerDisplayString()
+            public override string ToString()
             {
                 var sb = new StringBuilder();
 
-                sb.Append($"ReferenceCount: {SortedChildren.Count}, {Project.DebuggerDisplayString()}");
+                sb.Append($"ReferenceCount: {SortedChildren.Count}, {Project}");
 
                 return sb.ToString();
             }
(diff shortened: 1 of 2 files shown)


src/StructuredLogViewer.Core/ProjectGraph/RuntimeGraph.cs
```diff
@@ -11,6 +13,7 @@ namespace StructuredLogViewer.Core.ProjectGraph
 {
     public class RuntimeGraph
     {
+        [DebuggerDisplay("{DebuggerDisplayString()}")]
```
KirillOsenkov: Can you override ToString()?

It can be used in more situations such as logging, tests, etc. Also DebuggerDisplay makes the debugger eval a tiny bit slower.

src/StructuredLogViewer.Core/ProjectGraph/RuntimeGraph.cs
```diff
@@ -40,15 +43,37 @@ public RuntimeGraphNode(Project p)
...
+                var projectStartTime = GetNearestNonConflictingDateTime(child.Project.StartTime, sortedChildren);
+
+                sortedChildren.Add(projectStartTime, child);
                 sortedChildrenCached = null;
+
+                DateTime GetNearestNonConflictingDateTime(DateTime newKey, SortedList<DateTime, RuntimeGraphNode> collection)
+                {
+                    while (collection.ContainsKey(newKey))
+                    {
+                        newKey = new DateTime(newKey.Ticks + 1);
```
akoeplinger: this would preserve time zone info (not sure that's relevant here):

```suggestion
                        newKey = newKey.AddTicks(1);
```

src/StructuredLogViewer.Core/ProjectGraph/RuntimeGraph.cs
cdmihai: I'd rather the debugger be slower, than me having to click through various fields to understand what that object is. I'll rename it to ToString.

src/StructuredLogViewer.Core/ProjectGraph/RuntimeGraph.cs
```diff
@@ -11,6 +13,7 @@ namespace StructuredLogViewer.Core.ProjectGraph
 {
     public class RuntimeGraph
     {
+        [DebuggerDisplay("{ToString()}")]
```
KirillOsenkov: No need for this line at all now ;)

KirillOsenkov: I'll fix later np

cdmihai: Ah, didn't know the Debugger calls ToString by default on the objects if they override it.

Issue #319, opened 2020-03-19
MarkSearchResults is expensive during search


KirillOsenkov:
Marking the search results in the tree turns out to be very expensive during search.

There are a few things we could do.

1. Add a setting to toggle it on and off and set the default off. I've seen people ask what these dots are for and while I agree that it may be useful in some cases it doesn't seem to be warranted to be on by default.

2. Each result already has the node object when it is found - we could mark it as a result then and there, and avoid calling VisitAllChildren on an entire tree and do a lookup per result.

@ltrzesniewski let me know if you have some time looking into this. If not, that's OK.

KirillOsenkov:
And here's a sample of how to add a setting:

https://github.com/KirillOsenkov/MSBuildStructuredLog/blob/a47cc39d090a55d117fbb1abd4410a80821eb125/src/StructuredLogViewer.Core/SettingsService.cs#L306-L327

KirillOsenkov:
Yeah it's pretty bad on large logs:

![image](https://user-images.githubusercontent.com/679326/77030736-da623b80-695c-11ea-845c-96b33cbb916d.png)

KirillOsenkov:
Yes, I'm afraid we're going to need an option and have it off by default.

ltrzesniewski:
I _think_ I have spotted some optimizations opportunities back when I was developing the feature, but I didn't want to change the code too much in order to keep the PR as simple as possible.

Now, I actually always rely on the dots to find my search results, I guess I'm biased but they're very useful IMO. If people don't know what they mean maybe adding a legend would be warranted?

Give me a few days, I'll see what I can do with this.

Just to make sure I reproduce this correctly, is tree virtualization disabled for in your example, or did you use a very large log? I'm asking since I don't think I've ever experienced search performance issues on the logs I work with.

KirillOsenkov:
I have to support log sizes of 150MB and up :) Was testing on 37MB and 142MB. The biggest I have on my box is 480MB.

ltrzesniewski:
Actually compiling the Roslyn project gives a nice 110 MB binlog. When searching for the string "build" in that binlog, the search takes roughly 11 seconds on my machine.

I reduced that to 7s for now. It takes 5s when `MarkSearchResults` is commented out. I'm pretty sure I can still improve on that. 😉

KirillOsenkov:
Yes, I really want to optimize search as tightly as possible. 

One thing I was thinking of is parallelization.

[...]

PR #323, opened 2020-03-22
Use BaseNode instead of object

ltrzesniewski:
See discussion in #319.


KirillOsenkov:
Nice, I like it.

Historical reference: I spent a lot of time optimizing memory consumption of nodes, because large logs were taking up > 10 GB when loaded. I got rid of fields on nodes as much as possible, added flags to group booleans, etc. Some nodes didn't need to store a parent pointer, so there was a distinction between BaseNode and ParentedNode.

However with time as I was adding more functionality, more and more nodes needed to store a parent, so eventually all of them ended up being Parented ;) So your change makes sense now.

PR #330, opened 2020-04-11
moved property reassigment messages in new folder "properties" and subfolder with property name

AlexanderKapitanovski:
moved property reassigment messages in new folder "properties" and subfolder with property name


KirillOsenkov:
Thanks!

PR #332, opened 2020-04-14
added to search  $property name / value 

AlexanderKapitanovski:
Add $property name= and $property value= to only search property names or values respectively or  borth in the same time (names and values)

### DIFF
--- a/src/StructuredLogViewer.Core/NodeQueryMatcher.cs
+++ b/src/StructuredLogViewer.Core/NodeQueryMatcher.cs
@@ -21,6 +21,11 @@ namespace StructuredLogViewer
         public bool IncludeDuration { get; set; }
         public TimeSpan PrecalculationDuration { get; set; }
 
+        private bool searchInName { get; set; }
+        private bool searchInValue { get; set; }
+        private string nameToSearch { get; set; }
+        private string valueToSearch { get; set; }
+
         // avoid allocating this for every node
         [ThreadStatic]
         private static List<string> searchFieldsThreadStatic;
@@ -32,10 +37,6 @@ namespace StructuredLogViewer
             this.Query = query;
 
             this.Words = ParseIntoWords(query);
-            ////if (this.TypeKeywords == null)
-            ////    this.TypeKeywords = new List<string>();
-            string tag;
-            int eqaulindex;
 
             if (Words.Count == 1 &&
                 Words[0] is string potentialNodeIndex &&
@@ -50,6 +51,7 @@ namespace StructuredLogViewer
                     return;
                 }
             }
+
             for (int i = Words.Count - 1; i >= 0; i--)
             {
                 var word = Words[i];
@@ -60,7 +62,8 @@ namespace StructuredLogViewer
                     IncludeDuration = true;
                     continue;
                 }
-                if (word.Length > 2 && word[0] == '$' && word[1] != '(' &&(TypeKeyword == null || !TypeKeyword.Contains(word.Substring(1).ToLowerInvariant())))
+
+                if (word.Length > 2 && word[0] == '$' && word[1] != '(' && (TypeKeyword == null || !TypeKeyword.Contains(word.Substring(1).ToLowerInvariant())))
                 {
                     Words.RemoveAt(i);
                     TypeKeyword = word.Substring(1).ToLowerInvariant();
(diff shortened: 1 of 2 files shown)


src/StructuredLogViewer.Core/NodeQueryMatcher.cs
```diff
@@ -29,6 +32,10 @@ public NodeQueryMatcher(string query, IEnumerable<string> stringTable)
             this.Query = query;
 
             this.Words = ParseIntoWords(query);
+            ////if (this.TypeKeywords == null)
+            ////    this.TypeKeywords = new List<string>();
+            string tag;
+            int eqaulindex;
```
KirillOsenkov, line 38: `equalsIndex`

src/StructuredLogViewer.Core/NodeQueryMatcher.cs
```diff
@@ -29,6 +32,10 @@ public NodeQueryMatcher(string query, IEnumerable<string> stringTable)
             this.Query = query;
 
             this.Words = ParseIntoWords(query);
+            ////if (this.TypeKeywords == null)
+            ////    this.TypeKeywords = new List<string>();
```
KirillOsenkov, line 36: remove commented code

src/StructuredLogViewer.Core/NodeQueryMatcher.cs
```diff
@@ -43,7 +50,6 @@ public NodeQueryMatcher(string query, IEnumerable<string> stringTable)
                     return;
                 }
             }
-
```
KirillOsenkov, line 46: bring back the empty line - always an empty line after a `}` unless it's the last statement in the block

src/StructuredLogViewer.Core/NodeQueryMatcher.cs
```diff
@@ -54,8 +60,7 @@ public NodeQueryMatcher(string query, IEnumerable<string> stringTable)
                     IncludeDuration = true;
                     continue;
                 }
-
-                if (word.Length > 2 && word[0] == '$' && word[1] != '(' && TypeKeyword == null)
+                if (word.Length > 2 && word[0] == '$' && word[1] != '(' &&(TypeKeyword == null || !TypeKeyword.Contains(word.Substring(1).ToLowerInvariant())))
```
KirillOsenkov, line 63: space between `&&` and `(` (basically format the file using Ctrl+E,D)

src/StructuredLogViewer.Core/NodeQueryMatcher.cs
```diff
@@ -69,6 +74,19 @@ public NodeQueryMatcher(string query, IEnumerable<string> stringTable)
                     UnderMatcher = new NodeQueryMatcher(word, stringTable);
                     continue;
                 }
+                if ((word.StartsWith("name=", StringComparison.OrdinalIgnoreCase) || word.StartsWith("value=", StringComparison.OrdinalIgnoreCase)) && !word.EndsWith("="))
+                {
+                    eqaulindex = word.IndexOf("=");
```
KirillOsenkov, line 79: equalsIndex

src/StructuredLogViewer.Core/NodeQueryMatcher.cs
```diff
@@ -226,6 +244,7 @@ public static List<string> PopulateSearchFields(BaseNode node)
 
             if (node is NameValueNode nameValueNode)
             {
+
```
KirillOsenkov, line 247: remove empty line

src/StructuredLogViewer.Core/NodeQueryMatcher.cs
```diff
@@ -303,7 +322,9 @@ public SearchResult IsMatch(BaseNode node)
                     return null;
                 }
             }
-
+            
+            bool searchInName = AdditionalKeywords.ContainsKey("name");
```
KirillOsenkov, line 326: this lookup will be very expensive at runtime as it will run for 10,000,000 tree nodes. Can you declare the two booleans before the search starts?

src/StructuredLogViewer.Core/NodeQueryMatcher.cs
```diff
@@ -332,17 +353,38 @@ public SearchResult IsMatch(BaseNode node)
...
+                                    break;
+                                
+                            }
+                            if (searchInValue && j == 2 && word == AdditionalKeywords["value"])
+                            {                                
+                                    result.AddMatch(fullText, word);
+                                    anyFieldMatched = true;
+                                    break;                                
+                            }
+
```
KirillOsenkov, line 377: remove empty line after last closing `}`

src/StructuredLogViewer.Core/NodeQueryMatcher.cs
```diff
@@ -332,17 +353,38 @@ public SearchResult IsMatch(BaseNode node)
...
+                        if (node is NameValueNode && (searchInName ||  searchInValue))
+                        {
+                            if (searchInName && j == 1 && word == AdditionalKeywords["name"])
+                            {                                
+                                    result.AddMatch(fullText + (searchInValue? " =": ""), word, true);
+                                    anyFieldMatched = true;
+                                    break;
+                                
+                            }
+                            if (searchInValue && j == 2 && word == AdditionalKeywords["value"])
```
KirillOsenkov, line 371: insert an empty line after `}` if it's not the last statement ;)

src/StructuredLogViewer.Core/NodeQueryMatcher.cs
```diff
@@ -332,17 +353,38 @@ public SearchResult IsMatch(BaseNode node)
...
                         {
                             // if we matched a property in the name, show value as well since it's useful
                             fullText = named.ToString();
                         }
-
-                        result.AddMatch(fullText, word);
+                        // NameValueNode is special case have to check in witch field to search
+                        if (node is NameValueNode && (searchInName ||  searchInValue))
+                        {
+                            if (searchInName && j == 1 && word == AdditionalKeywords["name"])
```
KirillOsenkov, line 364: I think this will be expensive at runtime. Using a dictionary is very flexible but adds a lot of cost. Let's just hard code name and value strings.

AlexanderKapitanovski: I've cleaned up and added two strings and two boolean instead of dictionary, it's seems to be better.. . Thanks for your remarks :-)

src/StructuredLogViewer.Core/NodeQueryMatcher.cs
```diff
@@ -21,6 +21,11 @@ public class NodeQueryMatcher
         public bool IncludeDuration { get; set; }
         public TimeSpan PrecalculationDuration { get; set; }
 
+        private bool searchInName { get; set; }
+        private bool searchInValue { get; set; }
+        private string nameToSearch { get; set; }
```
KirillOsenkov, line 23: Do we really need the two bool properties? We can just use the logic of whether the string is null or not. But we can fix next time.

PR #333, opened 2020-04-15
Make imports a folder - fixes issue where TaskRunner didn't work when launched from the UI

mmitche:
When attempting to run a task from the UI, the log viewer uses a task
index which is computed as the tree is initially walked. This task index is
then passed onto the TaskRunner executable, which attempts to find the item with
the same index. The analyzers may add new nodes into the graph. Typically
this doesn't affect numbering because indexes are only computed for TimedNodes. But
the import analyzer uses a TimedNode for a folder, which affects numbering, leading
to the inability for the TaskRunner to find the task you want to execute.

Instead of using a TimedNode for the folder, use a Folder.

### DIFF
--- a/src/StructuredLogViewer.Core/Analyzers/ImportTreeAnalyzer.cs
+++ b/src/StructuredLogViewer.Core/Analyzers/ImportTreeAnalyzer.cs
@@ -74,12 +74,12 @@ namespace Microsoft.Build.Logging.StructuredLogger
                 return;
             }
 
-            if (rootProjectNode.Children.First() is TimedNode importsFolder && importsFolder.Name == Strings.Imports)
+            if (rootProjectNode.Children.First() is Folder importsFolder && importsFolder.Name == Strings.Imports)
             {
             }
             else
             {
-                importsFolder = new TimedNode()
+                importsFolder = new Folder()
                 {
                     Name = Strings.Imports
                 };


KirillOsenkov:
Unfortunately it's using a TimedNode on purpose, because they have IDs so you want to be able to "search under subtree". Folders don't have IDs so you can't search under a folder. And it's very useful to be able to search under the Imports folder.

Let me think of a different fix. I didn't realize that the numbering change when the tree is mutated presents a problem, but you're right, it is a problem.

Do you have a specific example where this breaks? A binlog perhaps that you could email me?

mmitche:
Got it. I originally though that the right way for this to be done is for the index to be computed at load time, but that would break the other scenario too.

KirillOsenkov:
@mmitche so here's what I'm thinking:
https://github.com/KirillOsenkov/MSBuildStructuredLog/commit/04f02a90f97d8c9d3b9d97c4543d8342e8fc8849

Would you try to build this change locally and see if this solves your problem?

Apologies for the bug, it must have been non-trivial to figure out the root cause.

KirillOsenkov:
I'll close this one but please let me know if my solution didn't work and we'll investigate further.

mmitche:
@KirillOsenkov I think that's a good fix. I thought about that first but wasn't sure if the expense of running the analyzers was worth it.

Btw, what are the chances we can get .NET Core support for the task runnner?

The primary problem now appears to be that the runner itself needs to be built for netcore, with modifications to assembly resolution to load any input dependenices.

KirillOsenkov:
If only I knew someone on the .NET team who could volunteer and help with this  ;)

Unfortunately I’m not an expert in .Net Core assembly loading etc and I really don’t have much free time nowadays. Ask around, maybe someone else wants to send a PR?

mmitche:
@KirillOsenkov I may be looking into it in the next couple months. It *may* be that it would require changes to the binlog format, but not sure.

KirillOsenkov:
@mmitche cool. Ping me if I can help with anything.

Another approach here is to generate a fake .proj file that just initializes the right properties and items and calls the task, then just debug actual MSBuild building this fake project and running this task. I thought about that but opted for a hand-written runner that emulates what MSBuild does.

KirillOsenkov:
@mmitche FYI I've just discovered that analyzers are not the only thing that impacts or mutates the tree.

There's also this setting:
image

Depending on this setting the targets will be either directly under the project or each target will be nested under the target that caused it to build.

KirillOsenkov:
I had to do this to respect the setting when reading the .binlog from TaskRunner:
https://github.com/KirillOsenkov/MSBuildStructuredLog/commit/f40387340e4a79b0578c747e8d4f5666e387c288

KirillOsenkov:
Published an update that hopefully fixes this:
https://github.com/KirillOsenkov/MSBuildStructuredLog/releases/tag/v2.1.117

PR #335, opened 2020-04-18
Implement notunder(...) syntax

AlexanderKapitanovski:
added notunder() syntax (multi node choose possible) and multi node choose to under()

### DIFF
--- a/src/StructuredLogViewer.Core/NodeQueryMatcher.cs
+++ b/src/StructuredLogViewer.Core/NodeQueryMatcher.cs
@@ -21,6 +21,11 @@ namespace StructuredLogViewer
         public bool IncludeDuration { get; set; }
         public TimeSpan PrecalculationDuration { get; set; }
 
+        private bool searchInName { get; set; }
+        private bool searchInValue { get; set; }
+        private string nameToSearch { get; set; }
+        private string valueToSearch { get; set; }
+
         // avoid allocating this for every node
         [ThreadStatic]
         private static List<string> searchFieldsThreadStatic;
@@ -32,10 +37,6 @@ namespace StructuredLogViewer
             this.Query = query;
 
             this.Words = ParseIntoWords(query);
-            ////if (this.TypeKeywords == null)
-            ////    this.TypeKeywords = new List<string>();
-            string tag;
-            int eqaulindex;
 
             if (Words.Count == 1 &&
                 Words[0] is string potentialNodeIndex &&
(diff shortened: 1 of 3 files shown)


src/StructuredLogViewer/Controls/BuildControl.xaml.cs
```diff
@@ -914,6 +919,15 @@ public void SearchInSubtree()
             }
         }
 
+        public void ExcludeSubtreeFromSearch()
+        {
+            var treeNode = treeView.SelectedItem as TimedNode;
+            if (treeNode != null)
```
KirillOsenkov, line 925: `if (treeView.SelectedItem is TimedNode treeNode)`

src/StructuredLogViewer.Core/NodeQueryMatcher.cs
```diff
@@ -69,9 +56,23 @@ public NodeQueryMatcher(string query, IEnumerable<string> stringTable)
...
                 {
-                    word = word.Substring(6, word.Length - 7);
+                    var nodeIndexText = word.Substring(7, word.Length - 8);
                     Words.RemoveAt(i);
-                    UnderMatcher = new NodeQueryMatcher(word, stringTable);
+                    if (int.TryParse(nodeIndexText, out var nodeIndex))
+                    {
+                        UnderNodes.Add(nodeIndex);
+                    }
+                    continue;
```
KirillOsenkov, line 65: we still need to have an UnderMatcher if the query is not a node index.

You can have queries like `$task Csc under($project Foo)`, so UnderMatcher recursively evaluates the expression in parentheses `$project Foo`

src/StructuredLogViewer.Core/NodeQueryMatcher.cs
```diff
@@ -22,6 +21,8 @@ public class NodeQueryMatcher
         private bool searchInValue { get; set; }
         private string nameToSearch { get; set; }
         private string valueToSearch { get; set; }
+        private HashSet<int> UnderNodes { get; set; } = new HashSet<int>();
```
KirillOsenkov, line 24: I think these will need to be converted to `List<NodeQueryMatcher>` to allow for arbitrary expressions in parentheses

AlexanderKapitanovski: sorry I didn't see this tipp $task Csc under($project Foo) and didn't imagine this way to search.  I will make it better.:-)

AlexanderKapitanovski: I've changed the logik to allow arbitrary expression in under() notunder(), need to look at this together

PR #342, opened 2020-04-25
added property value to result by search in property name

AlexanderKapitanovski:
added  property value to result by search in name and exlude property value from highlights

### DIFF
--- a/src/StructuredLogViewer.Core/ProxyNode.cs
+++ b/src/StructuredLogViewer.Core/ProxyNode.cs
@@ -69,9 +69,10 @@ namespace Microsoft.Build.Logging.StructuredLogger
                     Highlights.Add(fieldText.Substring(index, fieldText.Length - index));
                 }
 
-                if (result.SearchResultPaar.ContainsKey(fieldText))
+                string fieldValue = result.SearchResultPaar.Where(t => t.highlighted.Equals(fieldText)).Select(t => t.nothighleghted).FirstOrDefault();
+                if (fieldValue != null)
                 {
-                    Highlights.Add(" = " + TextUtilities.ShortenValue(result.SearchResultPaar[fieldText], "..."));
+                    Highlights.Add(" = " + TextUtilities.ShortenValue(fieldValue, "..."));
                 }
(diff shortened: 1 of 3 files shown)


src/StructuredLogViewer.Core/ProxyNode.cs
```diff
@@ -68,6 +68,11 @@ public void Populate(SearchResult result)
                 {
                     Highlights.Add(fieldText.Substring(index, fieldText.Length - index));
                 }
+
+                if (result.SearchResultPaar.ContainsKey(fieldText))
```
KirillOsenkov, line 72: `Pair` :)

src/StructuredLogViewer.Core/SearchResult.cs
```diff
@@ -8,6 +8,7 @@ public class SearchResult
     {
         public BaseNode Node { get; }
         public List<(string field, string match)> WordsInFields = new List<(string, string)>();
+        public Dictionary<string, string> SearchResultPaar = new Dictionary<string, string>();
```
KirillOsenkov, line 11: I'm a bit concerned with performance here. We can have a lot of results and allocating 96 bytes Dictionary for each seems a bit wasteful, since this is performance critical code. Maybe just keep two strings?

src/StructuredLogViewer.Core/SearchResult.cs
```diff
@@ -20,7 +21,7 @@ public SearchResult(BaseNode node, bool includeDuration = false)
             }
         }
 
-        public void AddMatch(string field, string word, bool addAtBeginning = false)
+        public void AddMatch(string field, string word, string notHighlightedtext = "", bool addAtBeginning = false)
```
KirillOsenkov, line 24: can you name it `notHighlightedText`?

KirillOsenkov:
If the match is in the property value, will it prepend the `name=` part?

I'm thinking for NameValuePairs could we instead just add a fake field that is `name = value` as a single match? This way the algorithm will highlight properly and there's no need for special cases.

PR #377, opened 2020-07-21
Show project entry targets in tree and tool tip

drewnoakes:
Makes it easier to see entry targets where there are many (such as in VS design-time builds).

image

### DIFF
--- a/src/StructuredLogger/ObjectModel/Project.cs
+++ b/src/StructuredLogger/ObjectModel/Project.cs
@@ -193,5 +193,42 @@ namespace Microsoft.Build.Logging.StructuredLogger
 
         public IReadOnlyList<string> EntryTargets { get; set; } = Array.Empty<string>();
         public IDictionary<string, string> GlobalProperties { get; set; } = ImmutableDictionary<string, string>.Empty;
+
+        public override string ToolTip
+        {
+            get
+            {
+                var sb = new StringBuilder();
+
+                sb.AppendLine(ProjectFile);
+
+                if (EntryTargets != null)
+                {
+                    sb.AppendLine();
+                    sb.AppendLine("Targets:");
+                    sb.AppendLine();
+                    foreach (var target in EntryTargets.OrderBy(target => target, StringComparer.InvariantCultureIgnoreCase))
+                    {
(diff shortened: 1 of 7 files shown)


src/StructuredLogger/ObjectModel/Project.cs
```diff
@@ -193,5 +193,42 @@ public bool IsLowRelevance
...
+                var sb = new StringBuilder();
+
+                sb.AppendLine(ProjectFile);
+
+                if (EntryTargets != null)
+                {
+                    sb.AppendLine();
+                    sb.AppendLine("Targets:");
+                    sb.AppendLine();
+                    foreach (var target in EntryTargets.OrderBy(target => target, StringComparer.InvariantCultureIgnoreCase))
```
drewnoakes, line 210: I ordered these alphabetically which makes it easier to read, but I fear it might hide information about target ordering. Should this sorting be removed?

KirillOsenkov: Shall we have a folder under the Project node for `Targets`? Just like the folder for Global Properties? Preserve the original order, and the user can sort if they want (any node's children can be sorted from the right-click context menu).

I'm torn on whether to sort the targets in the tooltip or not. Keep them sorted for now?

KirillOsenkov:
Let me know when it's ready, I'll merge.

drewnoakes:
> Shall we have a folder under the Project node for Targets? Just like the folder for Global Properties?

Great idea!

image

It'd be nice if these navigated to items below. I might come back to that in future.

Is it still worth showing them in the tree, alongside the project node?

KirillOsenkov:
Nice! Navigating to a target should be easy, see:
https://github.com/KirillOsenkov/MSBuildStructuredLog/blob/master/src/StructuredLogViewer/themes/Generic.xaml#L253
https://github.com/KirillOsenkov/MSBuildStructuredLog/blob/master/src/StructuredLogViewer/Controls/NodeHyperlinkControl.cs

drewnoakes:
Added a few more bits to these nodes:

image

I think this is good to go now.

PR #402, opened 2020-10-12
Make target's parent symbol depend on build reason

InKahootz:
This allows you to quickly tell what the reason target was the cause of the target.


InKahootz:
The tooltip was a great addition to the target but I found myself needing to know whether it was a before, after, or depends on.
Generally, the connecting symbol will point to the target that it caused it to run.

Let me know your thoughts on the current symbols b/c I know I'm changing the default direction. I just went with what I thought naturally first.

Here's a little image of it:
image

Thanks for your tool!

(If you could label this as 'hacktoberfest-accepted' if you think I contributed something that's worthy of a PR, that would be sweet too, but that's not a big deal)

PR #404, opened 2020-10-16
Mark Ctrl+F handle to not steal focus from search panel.

### LINKED ISSUE #289: Find in file places cursor in wrong textbox
When I am looking at a file and I do Ctrl + F to bring up the find box it places the cursor there for a brief flash and then the cursor moves to the search bar so I end up typing in the wrong spot.

paulomorgado:
Fixes #289


KirillOsenkov:
Thanks!

PR #405, opened 2020-10-18
Hold shift and use mouse wheel to scroll editor horizontally

drewnoakes:
Most other controls in the application supported this, but the text editor did not.


KirillOsenkov:
:+1:

PR #406, opened 2020-10-22
Browser: Add drag and drop support for loading file

akoeplinger:
Adds a drop target so you can simply drag and drop your binlog file into the browser window instead of needing to use the file picker:

binlog

/cc @KirillOsenkov


KirillOsenkov (approved):

KirillOsenkov:
Deployed! Thanks for the contribution!

PR #411, opened 2020-11-01
Targets netcoreapp3.1 too

jairbubbles:
- Allows the application to be launched with .NET Core runtime (along with .NET Framework runtime)
- Changed the target frameworks for `StructuredLogger` to exactly match the ones supported by `Microsoft.Build`
- Fix some `Process.Start` (there is a [breaking change](https://docs.microsoft.com/fr-fr/dotnet/core/compatibility/fx-core#change-in-default-value-of-useshellexecute) introduced with .NET Core)

### DIFF
--- a/src/StructuredLogger/StructuredLogger.csproj
+++ b/src/StructuredLogger/StructuredLogger.csproj
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
+	<TargetFrameworks>net472;netcoreapp2.1</TargetFrameworks>
     <NBGV_DoNotEmitNonVersionCustomAttributes>true</NBGV_DoNotEmitNonVersionCustomAttributes>
     <EmbedUntrackedSources>true</EmbedUntrackedSources>
   </PropertyGroup>
@@ -47,7 +47,7 @@
     </EmbeddedResource>
   </ItemGroup>
   <ItemGroup>
-    <None Include="$(IconFilePath)" Pack="true" PackagePath=""/>
+    <None Include="$(IconFilePath)" Pack="true" PackagePath="" />
   </ItemGroup>
   <PropertyGroup Condition=" '$(OS)' == 'Windows_NT' ">
     <StartAction>Program</StartAction>
--- a/src/StructuredLogViewer.Avalonia/Controls/TextViewerControl.xaml.cs
+++ b/src/StructuredLogViewer.Avalonia/Controls/TextViewerControl.xaml.cs
@@ -186,7 +186,7 @@ namespace StructuredLogViewer.Avalonia.Controls
                 filePath = SettingsService.WriteContentToTempFileAndGetPath(Text, extension);
             }
 
-            Process.Start(filePath);
+            Process.Start(new ProcessStartInfo(filePath){ UseShellExecute = true });
         }
 
         private void copyFullPath_Click(object sender, RoutedEventArgs e)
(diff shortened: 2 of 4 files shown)


src/StructuredLogger/StructuredLogger.csproj
```diff
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
```
KirillOsenkov, line 3: Curious why keeping this netstandard2.0 wasn't enough?

KirillOsenkov:
This looks good, if you undraft it I'll merge and fix up AppVeyor.

KirillOsenkov:
I still need to figure out how to redistribute a .NET Core app with Squirrel (the deployment engine I'm using). It may be trivial or not, I haven't investigated that. I'm thinking long term we should move to publishing to the Windows Store.

src/StructuredLogger/StructuredLogger.csproj
jairbubbles: To avoid warnings during nuget restore, since a few version MsBuild is not targetting netstandard2.0 anymore. I also have other pending changes where I update some NuGet references, I'll shoot another PR.

jairbubbles:
If you publish the .NET core as self-contained you should just have to include the output in the package and it should do the trick. I looked at the issues on the Squirrel repo and it seems some users managed to make it work even if the package itself is not .NET Core compliant. I looked a bit to make it compile in .NET Core too (I might create  a PR as well) but it's challenging as it's also using a lot of old dependencies.

src/StructuredLogger/StructuredLogger.csproj
jairbubbles: I'm speaking about thoses warnings:
```
warning NU1701: Package 'DeltaCompressionDotNet 1.0.0' was restored using '.NETFramework,Version=v4.6.1, .NETFramework,Version=v4.6.2, .NETFramework,Version=v4.7, .NETFramework,Version=v4.7.1, .NETFramework,Version=v4.7.2, .NETFramework,Version=v4.8' instead of the project target framework '.NETCoreApp,Version=v3.1'. This package may not be fully compatible with your project.
```

The most problematic is the Squirrel package.

KirillOsenkov:
I’m OK with Squirrel still running on .NET Framework if it does the job of delivering the payload.

src/StructuredLogger/StructuredLogger.csproj
KirillOsenkov: I guess a lot of users rely on just downloading StructuredLogger.dll and using it in their builds and it's convenient for them to have a single one in all cases.

Your fix makes sense, but I'll need to think about it more whether I'm ready to jump forward with targeting. Afraid to break consumers.

As a workaround you can add NoWarn to shut up NU1701.

jairbubbles:
@KirillOsenkov  I updated the PR. I saw that the avalonia version is using `net5.0`. Would make sense to use it here too.

ViktorHofer:
If you're thinking about self cook contained then I would suggest to use NET5 instead and possibly the linker option on it to reduce app size.

KirillOsenkov:
Sorry I never merged this and there have been changes to the projects since. I think at this point it's easier to abandon and start a new PR targeting net5.0 or net6.0. At least half of these changes are already in main.

jairbubbles:
Indeed, I created #463. This time I didn't touch the `StructuredLogger` lib even if I think it's a lie to tell it's `netstandard` compatible 😝

PR #423, opened 2020-12-06
Some more refactorings

Therzok:
This also fixes the build.


Therzok:
@KirillOsenkov fixed the bild issue, committed file is here.

PR #433, opened 2020-12-31
show snapshot testing on a user control

SimonCropp:
@KirillOsenkov thoughts? it allows u to visualize and verify how your forms/controls look without running the app


KirillOsenkov:
This is very neat, thanks!

Unfortunately adding tests right now is not on my priority list. Bigger fish to fry and I need to spend my limited attention in more impactful areas, such as making MSBuild faster and better for everyone.

But I do appreciate the sample, I'll be sure to use Verify when need arises.

PR #454, opened 2021-03-12
Simplify build

sharwell:
* Use WPF support provided by the SDK
* Update squirrel.windows from 1.4.0 to 1.4.4
* Address NU1701 and NU5128 warnings
* Simplify project files

@KirillOsenkov if there's anything you want removed from this group of changes let me know and I'll rebase to restrict it.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,7 +1,10 @@
 <Project>
 
   <PropertyGroup Label="Output directory">
-    <BaseIntermediateOutputPath>$(MSBuildThisFileDirectory)obj\$(MSBuildProjectName)\</BaseIntermediateOutputPath>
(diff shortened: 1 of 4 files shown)


KirillOsenkov:
Last time I tried to bump Squirrel I ran into a bug and had to rollback. The current version works fine and I'd rather not fix what isn't broken. Ideally of course I should go track down that squirrel bug and bump to latest when that is fixed, but I don't have time for that. Let's not bump Squirrel this time.

Other changes look fine.

sharwell:
> The current version works fine

Happy to remove from this PR, but I was not able to the full solution locally with 1.4.0 (some short of permission error involving open files).

KirillOsenkov:
Builds fine for me; what error are you getting? Send me the binlog

PR #455, opened 2021-03-12
Add search example for failed compiler server

### LINKED ISSUE #451: Summarize compiler server behavior
@jaredpar is updating VBCSCompiler to produce more actionable data relating to the use of the compiler server. For each compiler invocation, the following will be available:

* The compiler server is enabled, and was used for the compilation request (ideal case)
* The compiler server is disabled due to the project configuration (success case, but known to be slow)
* The compiler server is enabled, but was unable to serve the compilation request (failure case)
    * Failed to launch the compiler server
    * Server launched, but failed due to [cause]

Due to the performance impact of successful operation of the compiler server, we should offer a feature to summarize the results of all compilation requests present in a binary log for quick review.

sharwell:
Closes #451

### DIFF
--- a/src/StructuredLogViewer.Avalonia/Controls/BuildControl.xaml.cs
+++ b/src/StructuredLogViewer.Avalonia/Controls/BuildControl.xaml.cs
@@ -271,7 +271,8 @@ Right-clicking a project node may show the 'Preprocess' option if the version of
             "was not imported by ",
             "out-of-date",
             "csc $task",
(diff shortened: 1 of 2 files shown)


KirillOsenkov:
This is in the Avalonia project, you probably want this in the WPF project (as well/instead)

sharwell:
@KirillOsenkov Added to both now

KirillOsenkov (approved):

PR #476, opened 2021-04-24
Replace deprecated avalonia controls

rstm-sf:
Replace deprecated avalonia controls:
- DrawingPresenter: [Use Image control with DrawingImage source](https://github.com/AvaloniaUI/Avalonia/blob/d5985edd37521ec5d15b88eb230065ecb517291c/src/Avalonia.Controls/DrawingPresenter.cs#L8)
- DropDown: [Use ComboBox](https://github.com/AvaloniaUI/Avalonia/blob/d5985edd37521ec5d15b88eb230065ecb517291c/src/Avalonia.Controls/DropDown.cs#L7)


KirillOsenkov:
Thanks!

PR #477, opened 2021-04-24
Fix error when checking for the exist of msbuild

rstm-sf:
Fix error when checking for the exist of msbuild

And refactoring of Should-Not-Exist-File check

### DIFF
--- a/src/StructuredLogViewer.Avalonia/MainWindow.xaml.cs
+++ b/src/StructuredLogViewer.Avalonia/MainWindow.xaml.cs
@@ -408,7 +408,7 @@ namespace StructuredLogViewer.Avalonia
             openFileDialog.Filters.Add(new FileDialogFilter { Name = "Build Log (*.binlog;*.buildlog;*.xml)", Extensions = { "binlog", "buildlog", "xml" } });
             openFileDialog.Title = "Open a build log file";
             var result = await openFileDialog.ShowAndGetFileAsync(this);
-            if (result == null || !File.Exists(result))
+            if (!File.Exists(result))
             {
                 return;
             }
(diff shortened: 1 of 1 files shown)


src/StructuredLogViewer.Avalonia/MainWindow.xaml.cs
```diff
@@ -566,7 +566,7 @@ private async Task BrowseForMSBuildExe()
             };
 
             var fileName = await openFileDialog.ShowAndGetFileAsync(this);
-            if (fileName != null)
+            if (ShouldNotExistFile(fileName))
```
KirillOsenkov, line 569: Let's not introduce a function and just call `!File.Exists` here, also remove the null check as it will check for null.

src/StructuredLogViewer.Avalonia/MainWindow.xaml.cs
```diff
@@ -422,7 +422,7 @@ private async Task OpenProjectOrSolution()
             openFileDialog.Filters.Add(new FileDialogFilter { Name = "MSBuild projects and solutions (*.sln;*.*proj)", Extensions = { "sln", "*proj" } });
             openFileDialog.Title = "Open a solution or project";
             var result = await openFileDialog.ShowAndGetFileAsync(this);
-            if (result == null || !File.Exists(result))
+            if (ShouldNotExistFile(result))
```
KirillOsenkov, line 425: Let's remove the null check and just keep the `!File.Exists`

src/StructuredLogViewer.Avalonia/MainWindow.xaml.cs
rstm-sf: Ah, I agree. Thank you

KirillOsenkov (approved):
Nice, thanks

PR #506, opened 2021-06-12
Use a different suffix for line breaks in shortened text

ltrzesniewski:
Tasks can log lines with a line break at the end, and the viewer currently displays these like this:

image

This suggests there is more information in the message, but it's not true. This PR changes the display to this:

image

If there's only whitespace after the line break, it changes the suffix to "↵". This tells the user there's a line break, but there's nothing interesting afterwards.

Note that the space before the suffix belongs to the message in the example screenshots - the actual suffix is `"↵"`, not `" ↵"`.


KirillOsenkov:
Nice!

PR #511, opened 2021-06-16
Simplify calling ITask.Execute

0xced:
Making it easier to step into the `Execute` method with a debugger.

Also using a direct cast to `ITask` instead of `as ITask` because an `InvalidCastException` is easier to diagnose than a `NullReferenceException` several instructions later.


KirillOsenkov:
Huh, I don’t remember why I used reflection here.

0xced:
Yeah, that seemed unnecessary. Thanks for merging all my pull requests so quickly. 😃

PR #515, opened 2021-07-15
Fix reference paths handling in BinlogTool

ladipro:
I ran into crashes with the BinlogTool and a particular binlog which had references with absolute paths such as:

`/r:"D:\some\path\assembly.dll"`

This PR fixes it by:
- Removing the enclosing quotes.
- Mapping the reference path to the output file path (removes the colon before path-combining).

Also adding the missing `output_path` argument to the usage text.

### DIFF
--- a/src/BinlogTool/SaveFiles.cs
+++ b/src/BinlogTool/SaveFiles.cs
@@ -65,6 +65,7 @@ namespace BinlogTool
                     var reference = TryGetReference(argument);
                     if (reference != null)
                     {
+                        reference = ArchiveFile.CalculateArchivePath(reference);
                         var physicalReferencePath = GetPhysicalPath(outputDirectory, reference);
                         WriteEmptyAssembly(physicalReferencePath);
                     }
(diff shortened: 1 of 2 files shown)


src/BinlogTool/SaveFiles.cs
```diff
@@ -103,16 +104,30 @@ private void WriteEmptyAssembly(string physicalReferencePath)
...
+            else
+            {
+                return null;
             }
 
-            return null;
+            if (reference.Length >= 2)
+            {
+                // Remove enclosing quotes if present
+                if ((reference.StartsWith('\'') && reference.EndsWith('\'')) ||
```
KirillOsenkov, line 124: I have a TrimQuotes() extension method in TextUtilities.cs, not sure if it's available in this project, we should probably just add a linked file if it's not already linked.

KirillOsenkov (approved):

KirillOsenkov:
Thanks!

PR #516, opened 2021-07-24
Navigate into `Import` elements

ltrzesniewski:
Here's a new little feature:

image

You can now ctrl-click on an underlined `Import` element to open the file it points to.

An element won't be underlined if it points to different file paths in different evaluations, or if the file it points to is not available (because it never got imported for instance).

This could still be improved (Avalonia version, handling the back key, tooltips, selecting the target file if it's not unique, etc), but I didn't want to change too much at once.


KirillOsenkov:
Thanks!

ltrzesniewski:
After seeing a case, I think that ambiguous `Import` elements not being underlined is misleading, as it suggests no import was ever made. I'll send a fix later today that will open a menu that lets you choose one of the files.

PR #522, opened 2021-08-18
Bring API up-to-date with MSBuild 16.10

KirillOsenkov:
We had copies of types that are now available in MSBuild 16.10.

Delete our copies where they are identical to what's in the product.


KirillOsenkov:
@daveaglick code review please?

This should fix #521

daveaglick:
Reviewing after the fact, code looks great to me. Should greatly simplify my call sites - I ended up having to use an extern alias and code up two flows depending on if the event was coming from MSBuild or the StructureLogging reader (since they were different types). Now that the event args are binding to the same type I don't need to care where the event is coming from. 👍

PR #533, opened 2021-09-30
Compile as .NET 6, add arm64 support

filipnavara:
Since .NET 6 is right around the corner I thought I'd PR this up. Couple of people run into [the issue](https://github.com/dotnet/runtime/issues/59168) that they only have ARM64 .NET SDK installed on their system which makes the x64 .NET 5 apps not work out of the box due to apphost being x64 and finding only arm64 runtime. This is still [work in progress](https://github.com/dotnet/designs/blob/main/accepted/2021/x64-emulation-on-arm64/x64-emulation.md) from end user perspective on the .NET side.

This PR moves the target to .NET 6 and adds the ARM64 build. This makes the app work natively on Apple Silicon Macs.

### DIFF
--- a/src/StructuredLogViewer.Avalonia/StructuredLogViewer.Avalonia.csproj
+++ b/src/StructuredLogViewer.Avalonia/StructuredLogViewer.Avalonia.csproj
@@ -2,8 +2,8 @@
 
   <PropertyGroup>
     <OutputType>WinExe</OutputType>
-    <TargetFramework>net5.0</TargetFramework>
-    <RuntimeIdentifiers>win7-x64;ubuntu.14.04-x64;osx.10.12-x64</RuntimeIdentifiers>
+    <TargetFramework>net6.0</TargetFramework>
+    <RuntimeIdentifiers>win7-x64;ubuntu.14.04-x64;osx.10.12-x64;osx-arm64</RuntimeIdentifiers>
   </PropertyGroup>
   
   <ItemGroup>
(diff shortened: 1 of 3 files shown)


src/StructuredLogViewer.Avalonia/StructuredLogViewer.Avalonia.csproj
```diff
@@ -29,6 +29,8 @@
     <PackageReference Include="Avalonia.Diagnostics" Version="0.10.3" Condition=" '$(Configuration)' == 'Debug' " />
     <PackageReference Include="Avalonia.AvaloniaEdit" Version="0.10.0" />
     <PackageReference Include="System.ComponentModel.TypeConverter" Version="4.3.0" />
+    <PackageReference Include="SkiaSharp" Version="2.88.0-preview.145" />
+    <PackageReference Include="HarfBuzzSharp" Version="2.8.2-preview.140" />
```
akoeplinger, line 33: why do you need these additional package references?

filipnavara: At the time I wrote it the transitive references were not new enough to contain the osx-arm64 native assets.

akoeplinger: Might be worth a comment :)

src/StructuredLogViewer.Avalonia/StructuredLogViewer.Avalonia.csproj
```diff
@@ -29,6 +29,9 @@
     <PackageReference Include="Avalonia.Diagnostics" Version="0.10.3" Condition=" '$(Configuration)' == 'Debug' " />
     <PackageReference Include="Avalonia.AvaloniaEdit" Version="0.10.0" />
     <PackageReference Include="System.ComponentModel.TypeConverter" Version="4.3.0" />
+    <!-- Bump transtive references to preview versions that contain osx-arm64 native assets -->
```
KirillOsenkov, line 32: transtive is misspelled ;)

KirillOsenkov (approved):
:+1:

akoeplinger (approved):

PR #547, opened 2021-12-07
Handle ITaskItem parameter in task runner

0xced:
Having to convert a `ITaskItem[]` to a `ITaskItem` may happen with MSBuild target batching. For example, this happens with the [Microsoft.SourceLink.GitHub.GetSourceLinkUrl](https://github.com/dotnet/sourcelink/blob/1.1.1/src/SourceLink.GitHub/build/Microsoft.SourceLink.GitHub.targets#L37) task where the `SourceRoot` parameter is somehow translated from what should be a property with metadata to a list of task items.

Without this translation, the following error would occur:
> System.ArgumentException: Object of type 'Microsoft.Build.Framework.ITaskItem[]' cannot be converted to type 'Microsoft.Build.Framework.ITaskItem'.

### DIFF
--- a/src/TaskRunner/Executor.cs
+++ b/src/TaskRunner/Executor.cs
@@ -146,6 +146,16 @@ namespace TaskRunner
                 {
                     value = taskItems.Select(t => t.ItemSpec).ToArray();
                 }
+                else if (propertyInfo.PropertyType == typeof(ITaskItem))
+                {
+                    var item = taskItems.First();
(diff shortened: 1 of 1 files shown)


0xced:
Here's what the `SourceRoot` look like in the viewer for the `Microsoft.SourceLink.GitHub.GetSourceLinkUrl` task.

Screenshot of the SourceRoot as task items

KirillOsenkov (approved):

KirillOsenkov:
Thanks!

PR #549, opened 2021-12-14
Add Trace like graph

yuehuang010:
### Trace graph with x-axis as time.
image
 - Toggle showing Project Evaluation, Project, Targets, Tasks.  When there are too many items, it will just show Tasks.
image

 - Added a ruler to measure time in seconds.
 - Like timeline, click to highlight, hover to see more information, double to click to goto item in the log.

Fix timeline graph to show nodes in sorted order.
Fix timeline graph to all start at the same time.
Improved endpoint sorting so that stacking is more consistent.

### DIFF
--- a/src/StructuredLogViewer.Core/Timeline/Block.cs
+++ b/src/StructuredLogViewer.Core/Timeline/Block.cs
@@ -41,10 +41,73 @@ namespace StructuredLogViewer
         }
     }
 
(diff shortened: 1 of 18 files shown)


slang25:
This looks awesome!

KirillOsenkov (approved):

PR #554, opened 2022-01-13
Remove spaces from macOS artifacts

slang25:


KirillOsenkov (approved):

KirillOsenkov:
Thanks! 

I think we should also remove spaces from here too:
https://github.com/KirillOsenkov/MSBuildStructuredLog/blob/d42c0dbbe853f6ace4d0ad38ec35f05fa056e0b4/build-macos.cake#L13

Can you think of a reason why removing spaces there would break anything?

PR #564, opened 2022-02-04
Add publishing configuraiton for NativeAOT

kant2002:
When publish on NativeAOT applicaiton starts significantly faster.
I did not measure, but on old hardware I perceive this clearly.
Publishing can be done using
`dotnet publish -c Release -r linux-x64 -p:PublishNativeAot=True src/StructuredLogViewer.Avalonia/StructuredLogViewer.Avalonia.csproj`


KirillOsenkov:
mind adding a new file PublishNativeAOT.sh with the command?
`dotnet publish -c Release -r linux-x64 -p:PublishNativeAot=True src/StructuredLogViewer.Avalonia/StructuredLogViewer.Avalonia.csproj`

kant2002:
I add `cmd` as well. The final size of code is 144Mb for now. On Linux Native AOT put debug symbols to executable file, so it grow to 500Mb which would be percepted badly until all understand that. I will strip debug symbols on Linux for now.

KirillOsenkov:
Thanks!

PR #571, opened 2022-03-23
Log environment-derived properties per dotnet/msbuild#7484

Forgind:
We will need to update MSBuildPackageVersion once dotnet/msbuild#7484 merges.


src/StructuredLogger/Construction/Construction.cs
```diff
@@ -80,8 +80,15 @@ public void BuildStarted(object sender, BuildStartedEventArgs args)
                 lock (syncLock)
                 {
                     Build.StartTime = args.Timestamp;
-                    var properties = Build.GetOrCreateNodeWithName<Folder>(Intern(Strings.Environment));
-                    AddProperties(properties, args.BuildEnvironment);
+                    if (args.BuildEnvironment.Count > 0)
```
KirillOsenkov, line 83: need to check for null just in case

Forgind (commented):
We should be able to revert everything related to BuildFinishedEventArgs2 as well as the "EnvironmentDerivedProperties" string.

src/StructuredLogger/BinaryLogger/BinaryLogger.cs
```diff
@@ -55,7 +55,7 @@ public sealed class BinaryLogger : ILogger
         //   - TargetSkippedEventArgs: added OriginallySucceeded, Condition, EvaluatedCondition
         // version 14:
         //   - TargetSkippedEventArgs: added SkipReason, OriginalBuildEventContext
-        internal const int FileFormatVersion = 14;
+        internal const int FileFormatVersion = 15;
```
KirillOsenkov, line 58: no longer need to rev the format I think

src/StructuredLogViewer/Entrypoint.cs
```diff
@@ -10,6 +11,7 @@ public class Entrypoint
         [STAThread]
         public static void Main(string[] args)
         {
+            Debugger.Launch();
```
KirillOsenkov, line 14: probably don't want this change

Forgind: We don't want every MSBuildStructuredLogViewer user to try to debug it?? How else do we form the next generation of MSBuild devs? /s

KirillOsenkov:
Thanks!

PR #589, opened 2022-05-06
Move to net6.0

### LINKED ISSUE #586: Moving to net6.0
Several projects in the code base target `net5.0` which is going out of support in four days (May 10th).

https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core

Are there any reasons why we can't move the projects forward to `net6.0`? Happy to send a PR that does this but wanted to check before sending the PR.

jaredpar:
`net5.0` is going out of support in a few days so moved all of the
projects to target `net6.0` which is LTS

closes #586


KirillOsenkov (approved):

KirillOsenkov:
I pushed a fix for another occurrence directly to your branch, let's see if CI is happy now

PR #592, opened 2022-05-15
DuplicateWrites: Add support for robocopy task detection (from Microsoft.Build.Artifacts)

NickCraver:
Overall: adds RoboCopy task support for duplicate file copy detection. This is something very prevalent in some solutions I've been working on and would be a good win if we had it on the dupe detection front.

The task it's detecting is from the Microsoft.Build.Artifacts package which it turns out...isn't actually using Robocopy because...yeah, no clue. Here's the source of the task we're matching: https://github.com/microsoft/MSBuildSdks/blob/543e965191417dee65471ee57a6702289847b49b/src/Artifacts/Tasks/Robocopy.cs

### DIFF
--- a/src/StructuredLogger/Strings/Strings.cs
+++ b/src/StructuredLogger/Strings/Strings.cs
@@ -78,31 +78,32 @@ namespace Microsoft.Build.Logging.StructuredLogger
 
             TargetDoesNotExistBeforeTargetMessage = CreateRegex(GetString("TargetDoesNotExistBeforeTargetMessage"), 2);
 
-            string copyingFileFrom = GetString("Copy.FileComment");
-            string copyingFileFromEscaped = Escape(copyingFileFrom);
+            string copyingFileFromEscaped = Escape(GetString("Copy.FileComment"));
             CopyingFileFromRegex = new Regex(copyingFileFromEscaped
                 .Replace(@"\{0}", @"(?<From>[^\""]+)")
-                .Replace(@"\{1}", @"(?<To>[^\""]+)")
-                );
-
-            string robocopyingFileFrom = GetString("Robocopy.FileComment");
-            string robocopyingFileFromEscaped = Escape(robocopyingFileFrom);
-            RobocopyingFileFromRegex = new Regex(robocopyingFileFromEscaped
-                .Replace(@"\{0}", @"(?<From>[^\""]+)")
-                .Replace(@"\{1}", @"(?<To>[^\""]+)")
-                );
+                .Replace(@"\{1}", @"(?<To>[^\""]+)"), RegexOptions.Compiled);
 
             CreatingHardLinkRegex = new Regex(Escape(GetString("Copy.HardLinkComment"))
                 .Replace(@"\{0}", @"(?<From>[^\""]+)")
-                .Replace(@"\{1}", @"(?<To>[^\""]+)")
-                );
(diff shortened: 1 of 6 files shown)


KirillOsenkov:
I just added some help on how to re-run the ResourcesGenerator tool to regenerate Strings.json:
https://github.com/KirillOsenkov/MSBuildStructuredLog/commit/afe5eead3fc3e542b88b5827d0540cb011fab4c7

Basically you specify a list of resource strings from MSBuild you'd like to use here: 
https://github.com/KirillOsenkov/MSBuildStructuredLog/blob/afe5eead3fc3e542b88b5827d0540cb011fab4c7/src/ResourcesGenerator/ResourceCreator.cs#L22

run the ResourcesGenerator tool and it will generate Strings.json, so you'll be able to use it in Strings.cs

KirillOsenkov:
I'm not sure whether the Robocopy task is localized or not and how to retrieve the localized strings for it. If it's not localized you don't need to modify Strings.json at all and just hardcode the English string in Strings.cs. I'd be OK with just the English string for starters.

KirillOsenkov:
Looks great!

src/StructuredLogger/Strings/Strings.cs
```diff
@@ -78,24 +78,32 @@ private static void InitializeRegex()
...
                 .Replace(@"\{0}", @"(?<From>[^\""]+)")
-                .Replace(@"\{1}", @"(?<To>[^\""]+)")
-                );
+                .Replace(@"\{1}", @"(?<To>[^\""]+)"), RegexOptions.Compiled);
 
             CreatingHardLinkRegex = new Regex(Escape(GetString("Copy.HardLinkComment"))
                 .Replace(@"\{0}", @"(?<From>[^\""]+)")
-                .Replace(@"\{1}", @"(?<To>[^\""]+)")
-                );
+                .Replace(@"\{1}", @"(?<To>[^\""]+)"), RegexOptions.Compiled);
```
NickCraver, line 88: I noticed these weren't optimized (but run quite a bit on some builds) - happy to separate this optimization into another PR to keep things clean!

NickCraver:
Whelp, I learned a few things. The reason I couldn't find a localization, even internally is _that task isn't actually using Robocopy_. It's just a file loop - see the Microsoft.Build.Artifacts package source here: https://github.com/microsoft/MSBuildSdks/blob/main/src/Artifacts/Tasks/Robocopy.cs#L66-L70 

I still think this analyzer is useful and I've added the skipped and failed cases because they're all ultimately duplicate writes. e.g. a skip may be a date mismatch and non-deterministic for a build and a fail could be a attempted simultaneous double write. In either case I think it's in the spirit of what we're trying to detect. Happy to adjust though if that doesn't sound right!

cc @jaredpar who may be interested in this one.

src/StructuredLogger/Strings/Strings.cs
KirillOsenkov: This is fine, thanks.

I benchmarked the difference a while back and it does have benefits:
https://github.com/KirillOsenkov/Benchmarks/blob/29e6de831b11abd9485b16cc24f3eb310e989ed6/src/Tests/Regex.cs#L8-L29

KirillOsenkov:
Haha, Robocopy isn't using Robocopy! TIL.

KirillOsenkov (approved):

jaredpar:
Love it!

PR #608, opened 2022-06-30
Add CPP Details by parsing MTT and Bt+ into Tracing View

yuehuang010:
## Extract the diagnostic messages and add them to the trace view:
 - Extract the diagnostic messages from MultiToolTask for the inner build times.
 - Extract tool specific (CL /Bt+, LIB /TIME, and LINK /TIME) diagnostic messages.  Since these tools diagnostic messages are optional, detected for the switch to then suppressed MultiToolTask messages. 
 - Removed the unused "Others".

"Cpp Details" Off:
image

"Cpp Details" On:
image

## Add an option merge all the nodes into one node.  Screenshot of the updated context menu:
image

## Improve the loading of the graph by only rendering the initial viewport.  
 - As the user scrolls, add new elements are appended to the graph.
 - Now that loading is fast enough, there is no longer a need to apply a limit.


KirillOsenkov:
Let me know when this is ready for review and testing.

yuehuang010:
> Let me know when this is ready for review and testing.

@KirillOsenkov Ready Now.

src/StructuredLogger/ObjectModel/Message.cs
```diff
@@ -8,7 +8,7 @@ public class Message : TextNode, IHasRelevance, IHasSourceFile, IHasLineNumber
         /// <summary>
         /// Let's see if we even need timestamp, it's just eating memory for now
         /// </summary>
-        public DateTime Timestamp { get { return DateTime.MinValue; } set { } }
+        public DateTime Timestamp { get; set; }
```
KirillOsenkov, line 11: This is going to very significantly increase memory consumption. Are we sure we need this? Is there a way around it that doesn't need storage for every message?

src/StructuredLogViewer/Controls/TracingControl.xaml.cs
```diff
@@ -783,6 +858,16 @@ private static Brush ChooseBackground(Block block)
         private void ResetZoom_Click(object sender, RoutedEventArgs e)
         {
             zoomSlider.Value = 1;
+            UpdatedGraph(scrollViewer.VerticalOffset + scrollViewer.ViewportWidth);
+        }
+
+        private void ScrollViewer_ScrollChanged(object sender, ScrollChangedEventArgs e)
+        {
+            if (e.HorizontalChange == 0)
+                return;
```
KirillOsenkov, line 867: I should add a setting to .editorconfig to always require braces around the if statement body.

src/StructuredLogViewer/Controls/BuildControl.xaml.cs
```diff
@@ -386,7 +386,7 @@ private void PopulateTimeline()
         {
             if (this.timeline.Timeline == null)
             {
-                var timeline = new Timeline(Build);
+                var timeline = new Timeline(Build, false);
```
KirillOsenkov, line 389: let's add the named argument: `includeCpp: false`

In general I prefer to include named arguments when passing literals such as `true`, `false` and `null`.

src/StructuredLogViewer.Core/Timeline/Timeline.cs
```diff
@@ -43,11 +109,254 @@ private void Populate(Build build)
...
+                {
+                    IEnumerable<Block> blocks = PopulateCppNodes(node);
+                    lane.AddRange(blocks);
+                }
+
+                lane.Add(CreateBlock(node));
             });
         }
 
+        private IEnumerable<Block> PopulateCppNodes(TimedNode node)
```
KirillOsenkov, line 122: Logically this type of work belongs in an analyzer. After a build is read from a binlog, it is analyzed, and additional (virtual) nodes are added to the tree. See for example:
https://github.com/KirillOsenkov/MSBuildStructuredLog/blob/f1a3384d5e883432b410788a86f454b65499ec25/src/StructuredLogger/Analyzers/BuildAnalyzer.cs#L318

KirillOsenkov: I'm OK with leaving this logic here, but if it makes sense, consider moving it to a CppAnalyzer.

src/StructuredLogViewer.Core/Timeline/Timeline.cs
```diff
@@ -1,31 +1,97 @@
...
             build.VisitAllChildren<TimedNode>(node =>
             {
                 if (node is Build)
                 {
+                    if (includeCpp)
+                    {
+                        // Search for Global /Bt+ and /TIME
+                        node.FindFirstChild<Folder>(childNode =>
+                        {
+                            if (childNode.Name == "Environment")
```
KirillOsenkov, line 72: use Strings.Environment

src/StructuredLogViewer.Core/Timeline/Timeline.cs
```diff
@@ -1,31 +1,97 @@
...
 namespace StructuredLogViewer
 {
     public class Timeline
     {
-        private Build build;
+        // parse for:
+        // midl.exe will run on 8 out of 8 file(s) in 8 batches.  Startup phase took 46.0028ms.
+        // Task 'CodeStore.idl' took 2244ms.
+        // Cleanup phase took 68ms.
+        const string regexStartupPhase = @"^.*\.exe will run on (?'activeFiles'[0-9]*) out of (?'totalFiles'[0-9]*) file\(s\) in [0-9]* batches\.\s*Startup phase took (?'msTime'[0-9]*.[0-9]*)ms.$";
```
KirillOsenkov, line 19: Consider moving these regexes to Strings.cs (it has examples of other regexes already)

src/StructuredLogViewer.Core/Timeline/Lane.cs
```diff
@@ -8,18 +9,29 @@ public class Lane
...
 
         public void Add(Block block)
+        {
+            if (IsValid(block))
+                Blocks.Add(block);
+        }
+
+        public void AddRange(IEnumerable<Block> blocks)
+        {
+            Blocks.AddRange(blocks.Where(b => IsValid(b)));
```
KirillOsenkov, line 19: slightly worried about perf here. Is this a hotspot? Perhaps worth rewriting as a simple foreach loop and call Blocks.Add?

src/StructuredLogViewer.Core/Timeline/Lane.cs
```diff
@@ -8,18 +9,29 @@ public class Lane
         public List<Block> Blocks { get; set; } = new List<Block>();
 
         public void Add(Block block)
+        {
+            if (IsValid(block))
+                Blocks.Add(block);
```
KirillOsenkov, line 14: let's surround with braces

src/StructuredLogViewer.Core/Timeline/Timeline.cs
```diff
@@ -1,31 +1,97 @@
...
         }
 
-        private void Populate(Build build)
+        private void Populate(Build build, bool includeCpp = false)
         {
             build.VisitAllChildren<TimedNode>(node =>
             {
                 if (node is Build)
                 {
+                    if (includeCpp)
```
KirillOsenkov, line 67: If possible, I'd like to avoid Cpp-specific logic in Timeline.cs. Let's consider moving it to the Analyzers folder.

src/StructuredLogger/ObjectModel/Message.cs
yuehuang010: This is going to be challenging, I need to look around.

KirillOsenkov: Maybe make a `Dictionary<Message, DateTime>`?

I tightly optimized the memory consumption of the core data structures so the viewer can open huge binlogs with minimal memory consumption. In large binlogs there can be millions of messages, so not storing extra 8 bytes on each message was a significant win.

But if it's really difficult, we can keep your change.

src/StructuredLogViewer.Core/Timeline/Timeline.cs
yuehuang010: Created CppAnalyzer to hold all the cpp stuff.

src/StructuredLogViewer.Core/Timeline/Timeline.cs
yuehuang010: Moved to CppAnalyzer.

src/StructuredLogViewer.Core/Timeline/Lane.cs
yuehuang010: Removed as it is unused.

src/StructuredLogViewer.Core/Timeline/Timeline.cs
yuehuang010: Moved.

src/StructuredLogger/ObjectModel/Message.cs
yuehuang010: Still looking.

src/StructuredLogger/ObjectModel/Message.cs
```diff
@@ -3,12 +3,20 @@
 
 namespace Microsoft.Build.Logging.StructuredLogger
 {
+    public class TimedMessage : Message
```
KirillOsenkov, line 6: Great idea!

KirillOsenkov (approved):
Let me know when this iteration is ready.

PR #619, opened 2022-09-06
Change truncation strategy for environment variables

Forgind:
This shifts to using MSBUILDLOGALLENVIRONMENTVARIABLES to determine if we logged the full environment or not and gives a useful message if we truncated it.

### DIFF
--- a/src/StructuredLogger/BinaryLogger/BinaryLogger.cs
+++ b/src/StructuredLogger/BinaryLogger/BinaryLogger.cs
@@ -55,7 +55,7 @@ namespace Microsoft.Build.Logging.StructuredLogger
         //   - TargetSkippedEventArgs: added OriginallySucceeded, Condition, EvaluatedCondition
         // version 14:
         //   - TargetSkippedEventArgs: added SkipReason, OriginalBuildEventContext
-        internal const int FileFormatVersion = 14;
+        internal const int FileFormatVersion = 15;
 
         private Stream stream;
         private BinaryWriter binaryWriter;
--- a/src/StructuredLogger/BinaryLogger/BinaryLogger.cs
+++ b/src/StructuredLogger/BinaryLogger/BinaryLogger.cs
@@ -55,7 +55,7 @@ namespace Microsoft.Build.Logging.StructuredLogger
         //   - TargetSkippedEventArgs: added OriginallySucceeded, Condition, EvaluatedCondition
         // version 14:
         //   - TargetSkippedEventArgs: added SkipReason, OriginalBuildEventContext
-        internal const int FileFormatVersion = 15;
+        internal const int FileFormatVersion = 14;
 
         private Stream stream;
         private BinaryWriter binaryWriter;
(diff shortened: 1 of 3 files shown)


src/StructuredLogger/BinaryLogger/BinaryLogger.cs
```diff
@@ -55,7 +55,7 @@ public sealed class BinaryLogger : ILogger
         //   - TargetSkippedEventArgs: added OriginallySucceeded, Condition, EvaluatedCondition
         // version 14:
         //   - TargetSkippedEventArgs: added SkipReason, OriginalBuildEventContext
-        internal const int FileFormatVersion = 14;
+        internal const int FileFormatVersion = 15;
```
KirillOsenkov, line 58: Don't think we need to rev the file format version here.

We need to ensure that we keep the file format version in sync between MSBuild and this repo. I guess the file format doesn't really change. Let's see if we find another way.

KirillOsenkov: Also this wouldn't really help anyway because the binlog files are recorded by MSBuild and not by this repo, so changing this repo's version would only impact binlogs recorded by this logger (which is 0.1% of cases maybe).

src/StructuredLogger/Construction/Construction.cs
```diff
@@ -95,15 +95,23 @@ public void BuildStarted(object sender, BuildStartedEventArgs args)
...
                     {
+                        if (args.BuildEnvironment?.Count > 0)
+                        {
+                            AddProperties(EnvironmentFolder, args.BuildEnvironment);
+                        }
+
                         EnvironmentFolder.AddChild(new Note
                         {
-                            Text = Intern(Strings.NoEnvironment)
+                            Text = Intern(
```
KirillOsenkov, line 111: I'm thinking that we don't really need to change anything but the string, if we craft if carefully (and we can reuse the string for NoEnvironment, so no need to introduce a new string for TruncatedEnvironment).

How about:

```
Starting with MSBuild 17.4, only some environment variables are included in the log: the ones read during the build and the ones prefixed with MSBUILD, DOTNET_ or COMPLUS_. Define MSBUILDLOGALLENVIRONMENTVARIABLES to log all environment variables during the build.
```

And leave the previous logic intact (if the dictionary contains items, add them to the tree)

KirillOsenkov: and just always add the note to the tree, in all cases

Forgind: Good point; hadn't been thinking both changes would be in 17.4. I'll do that then, thanks!

Forgind: I think it still makes sense to only conditionally add the note—as an indicator to us that the user is logging all environment variables or not if nothing else.

KirillOsenkov: I think it’s pretty obvious anyway from the list itself.

We do log the exact MSBuild version that the binlog was created from, so we could do conditional based on that. However that feels flimsy.

KirillOsenkov (approved):

PR #620, opened 2022-09-19
Add Platform/Configuration Adornment, CppAnalyzer QOL improvements

yuehuang010:
### Add Platform/Configuration Adornment
Add platform and configuration into the existing TargetFramework property for project nodes.  Platform and Configuration are the primary properties and is often a cause of issue when solution are incorrectly batched.
<details>
<summary>Screen Shot</summary>

</details>

### CppAnalyzer QOL improvements
- Turned off Cpp Details by default.
- Fix an issue where not all Cpp Details are active for all task.  Now each task has a boolean to indicate when it is present.  IE. CL.exe can be have details but Linker doesn't.
- Merge C1/C2 messages from BT Time, cutting the number of nodes in half.
- Align the Compile Time to the right (previously center).  The start time is unknown but the end time is almost guarantee to be correct.
- Add more rounding to MTT messages to make the graph fit better.

### DIFF
--- a/src/StructuredLogger/Analyzers/BuildAnalyzer.cs
+++ b/src/StructuredLogger/Analyzers/BuildAnalyzer.cs
@@ -2,20 +2,21 @@
 using System.Collections.Generic;
 using System.Linq;
 using StructuredLogViewer;
+using static Microsoft.Build.Logging.StructuredLogger.CppAnalyzer;
 
 namespace Microsoft.Build.Logging.StructuredLogger
 {
     public class BuildAnalyzer
     {
-        private Build build;
-        private DoubleWritesAnalyzer doubleWritesAnalyzer;
-        private ResolveAssemblyReferenceAnalyzer resolveAssemblyReferenceAnalyzer;
-        private CppAnalyzer cppAnalyzer;
-        private int index;
-        private Dictionary<string, (TimeSpan TotalDuration, Dictionary<string, TimeSpan> ParentDurations)> taskDurations
+        private readonly Build build;
+        private readonly DoubleWritesAnalyzer doubleWritesAnalyzer;
+        private readonly ResolveAssemblyReferenceAnalyzer resolveAssemblyReferenceAnalyzer;
+        private readonly CppAnalyzer cppAnalyzer;
+        private readonly Dictionary<string, (TimeSpan TotalDuration, Dictionary<string, TimeSpan> ParentDurations)> taskDurations
             = new Dictionary<string, (TimeSpan TotalDuration, Dictionary<string, TimeSpan> ParentDurations)>();
         private readonly List<Folder> analyzerReports = new List<Folder>();
         private readonly List<Folder> generatorReports = new List<Folder>();
+        private int index;
 
         public BuildAnalyzer(Build build)
         {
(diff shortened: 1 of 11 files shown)


src/StructuredLogger/Analyzers/BuildAnalyzer.cs
```diff
@@ -311,6 +315,8 @@ private void PostAnalyzeProject(Project project)
             }
         }
 
+
+
```
KirillOsenkov, line 319: Let's remove these empty lines

src/StructuredLogger/ObjectModel/IProjectOrEvaluation.cs
```diff
@@ -1,7 +1,77 @@
...
+using System.Collections;
+using System.Collections.Generic;
+using System.Configuration;
+
 namespace Microsoft.Build.Logging.StructuredLogger
 {
     public interface IProjectOrEvaluation
     {
-        string TargetFramework { get; set; }
+        public string AdornmentString { get; }
```
KirillOsenkov, line 10: no need for the public keyword

src/StructuredLogger/ObjectModel/IProjectOrEvaluation.cs
```diff
@@ -1,7 +1,77 @@
...
+        private class IProjectOrEvaluationComparer : IEqualityComparer<IProjectOrEvaluation>
+        {
+            public bool Equals(IProjectOrEvaluation x, IProjectOrEvaluation y)
+            {
+                return x.TargetFramework == y.TargetFramework
+                    && x.Platform == y.Platform
+                    && x.Configuration == y.Configuration;
+            }
+
+            public int GetHashCode(IProjectOrEvaluation obj)
```
KirillOsenkov, line 30: You can just take advantage of the ValueTuple hash:
public int GetHashCode(IProjectOrEvaluation obj) => (TargetFramework, Platform, Configuration).GetHashCode();

src/StructuredLogger/ObjectModel/IProjectOrEvaluation.cs
```diff
@@ -1,7 +1,77 @@
...
+                int hashcode = 123456789;
+                hashcode = hashcode * 987654321 ^ (obj.TargetFramework == null ? 1 : obj.TargetFramework.GetHashCode());
+                hashcode = hashcode * 987654321 ^ (obj.Platform == null ? 1 : obj.Platform.GetHashCode());
+                hashcode = hashcode * 987654321 ^ (obj.Configuration == null ? 1 : obj.Configuration.GetHashCode());
+
+                return hashcode;
+            }
+        }
+
+        private const string seperator = ",";
```
KirillOsenkov, line 41: spelling: needs to be "separator" instead of "seperator"

src/StructuredLogger/ObjectModel/IProjectOrEvaluation.cs
```diff
@@ -1,7 +1,77 @@
...
+
+        private static IProjectOrEvaluationComparer comparer = new();
+
+        private static Dictionary<IProjectOrEvaluation, string> AdornmentStringCache = new(comparer);
+
+        public static string GetAdornmentString(this IProjectOrEvaluation proj)
```
KirillOsenkov, line 47: Not sure if the cache is the best solution here. How about you add AdornmentString on all implementations of IProjectOrEvaluation, and just have a property there that is calculated from the other three properties on demand? We could cache it there or calculate on demand every time.

src/StructuredLogger/ObjectModel/IProjectOrEvaluation.cs
yuehuang010: Thanks, that worked.  My concern was dealing with nulls and empty string.

src/StructuredLogger/ObjectModel/IProjectOrEvaluation.cs
yuehuang010: I am not 100% sure what you had in mind.  The cache does reduce the total number of strings as every project will have a new string for every view.

KirillOsenkov: Ah, sorry, I misunderstood how your cache works. I like it. All good.

PR #628, opened 2022-11-05
TraceView: Add option to show P2P relation when a project is highlighted

yuehuang010:
## Add a feature to show P2P relationship when a project node is selected.
- Remove some common P2P targets to avoid clutter.  ~10% of total targets, but huge visual real estate because they are long running.
- To compensate for the loss of details, extend highlight to show parent and dependent projects.
- Refactor Highlight mechanic to an single overlay Canvas.
- Removed ContentControl() to speed up loading.  Est. ~10% faster.
- Moved some context menu into sub-menu as the menu is starting to get long.
- As part of the refactor, improved hit rate of GoToTimedNode() by reloading all TextBlocks.

Misc.
- Avoid failing to load C++ Analysis when a duplicated message is printed.  Not sure why such message is printed, but best not to fail.

<details><summary>After</summary>

</details>

<details><summary>Before</summary>

</details>


KirillOsenkov (approved):
Looks great!

PR #642, opened 2022-12-19
Make icons a bit prettier

### LINKED ISSUE #598: Styles for Live Blazor website
I notice some glitches in UI on Desktop in the tree view, so I take liberty and play with Chrome to improve status

This small snippet should vertically align icon in the tree with text representing the tree.
```
.ui-treenode-content div {
    vertical-align: middle;
}

img {
    padding: 0 6px 3px 0;
}
```
If you add this style `margin-bottom: 0.5rem;` directly to `Dark Mode` button, I think gap from bottom and from top would be same, and consistent with other 3 buttons.

I see that icons in the tree do not aligned between nodes with children, and child-less nodes. Have no fix for that.

kant2002:
**before**
image

**after**
image

Closes #598

### DIFF
--- a/Blazor App/Shared/NavMenu.razor
+++ b/Blazor App/Shared/NavMenu.razor
@@ -1,4 +1,4 @@
-﻿@using StructuredLogViewerWASM.Pages
+@using StructuredLogViewerWASM.Pages
 @using System.IO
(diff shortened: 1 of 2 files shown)


KirillOsenkov (approved):

KirillOsenkov:
Thanks!

PR #650, opened 2023-01-27
Optimize NuGet package dependencies.

### LINKED ISSUE #649: Consider removing the `Microsoft.Build` package dependency, and the `TargetGraph` class.
The `StructuredLogger.MSBuild` package depends on `Microsoft.Build` which is incompatible with .NET Standard 2.0 ("works" via the compatibility shim), weighs almost 4MB and brings quite many dependencies. It would be nice if we stopped depending on it.

Fortunately the only place where it is used is in [`TargetGraph`](https://github.com/KirillOsenkov/MSBuildStructuredLog/blob/main/src/StructuredLogger/Construction/TargetGraph.cs), a class with relatively simple logic that is nowhere else used in this repro. Depending on your assessment of its importance and your appetite for breaking changes, we can:

* Just remove it.
* Obsolete it and remove it after a couple months/releases.
* Do nothing.

(and thank you very much for this project, it has allowed me to pierce inside `dotnet build`'s inner workings and helped me do lots of stuff with MSBuild!)

teo-tsirpanis:
Fixes #649.

### DIFF
--- a/src/StructuredLogViewer/StructuredLogViewer.csproj
+++ b/src/StructuredLogViewer/StructuredLogViewer.csproj
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <TargetFrameworks>net472;net6.0-windows</TargetFrameworks>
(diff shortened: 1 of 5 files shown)


KirillOsenkov (approved):

KirillOsenkov:
Hmm, wait, I'm getting a build failure:

image

The viewer project (StructuredLogViewer.csproj) does depend on Microsoft.Build.Tasks.Core.dll because it ships it with the app:
https://github.com/KirillOsenkov/MSBuildStructuredLog/blob/cdccacae9c280de7c60823e746dbc99309240aee/src/StructuredLogViewer/MSBuildStructuredLogViewer.nuspec#L34

I think we'll need to add the Tasks.Core PackageReference to StructuredLogViewer.csproj

KirillOsenkov (changes requested):
Need to add a PackageReference to Tasks to StructuredLogViewer.csproj so that the build succeeds

teo-tsirpanis:
CI is green now.

KirillOsenkov (approved):
:+1:

PR #653, opened 2023-02-07
Add support for logging assembly loads logging

JanKrivanek:
Related to: https://github.com/dotnet/msbuild/issues/6303
Deature in MSBuild: https://github.com/dotnet/msbuild/pull/8316

### Context

Adding support for showing assembly load events, prduced by the MSBuild.
There is no special treating of the events - they get disaplayed as text message. For sample messages, see the MSBuild PR listed at top.
Sample viewer screenshot:

image

### DIFF
--- a/src/StructuredLogger/AssemblyLoadBuildEventArgs2.cs
+++ b/src/StructuredLogger/AssemblyLoadBuildEventArgs2.cs
@@ -1,0 +1,67 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+using Microsoft.Build.Framework;
+using Microsoft.Build.Shared;
+
+// TODO: remove whole file once AssemblyLoadBuildEventArgs are imported via Microsoft.Build.Framework package
+
+namespace Microsoft.Build.Framework
+{
+    internal enum AssemblyLoadingContext
+    {
+        TaskRun,
+        Evaluation,
+        SdkResolution,
+        LoggerInitialization,
+    }
+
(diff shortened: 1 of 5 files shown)


KirillOsenkov (approved):

JanKrivanek:
@KirillOsenkov Do you have any quick thoughts why the same event is sometimes displayed as Message sometimes as Property?

image

The Property display happens for `MSBuild` tasks (indicated by `BuildEventCotnext`).

Unless this is expected or unless there might be some obvious reason for that I'll of cource investigate closer (I just wanted to see if I'm not missing anything)

KirillOsenkov:
it looks for space, equals, space to guess whether it’s a property :) The Version = string sometimes contains spaces and sometimes doesn’t :) We can tune the heuristic to account for this case. It ‘s in MessageProcessor.cs somewhere.

JanKrivanek:
Thank you @KirillOsenkov!
It got me into the right direction - the fix proposal: https://github.com/KirillOsenkov/MSBuildStructuredLog/pull/654

Something that seems strange to me now is that those messages do not appear in the search results:

image

Do you have any idea for reason for that?

KirillOsenkov:
yes you need to intern the strings in a special way to include them in search, look for Intern() methods used everywhere in Construction.cs and MessageProcessor.cs

JanKrivanek:
Thank you!
Works like a charm now! ;-) (https://github.com/KirillOsenkov/MSBuildStructuredLog/pull/654)

PR #654, opened 2023-02-09
Treat Assembly Loads as messages

JanKrivanek:
### Context

'Assembly loaded' messages were interpreted as Properties on `MSBuild` target.
https://github.com/KirillOsenkov/MSBuildStructuredLog/pull/653#issuecomment-1423025459

### Fix

Special case args by `AssemblyLoadedBuildEventArgs` type before special case handling `MSBuild` target


KirillOsenkov (approved):
:+1:

PR #655, opened 2023-02-14
Add Cpp Project Icon

yuehuang010:
Add double click on Meta Project message that will open the metaproject file.
Fix missing vb and fs project icons from search results.
Add Cpp Project Icon.
image

### DIFF
--- a/src/StructuredLogger/Search/ProxyNode.cs
+++ b/src/StructuredLogger/Search/ProxyNode.cs
@@ -236,20 +236,15 @@ namespace Microsoft.Build.Logging.StructuredLogger
 
         private string GetProjectFileExtension()
         {
-            string result = null;
+            string result = "other";
 
-            if (Original is Project project)
+            if (Original is Project project && !string.IsNullOrEmpty(project.ProjectFileExtension))
             {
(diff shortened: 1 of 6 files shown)


src/StructuredLogger/Search/ProxyNode.cs
```diff
@@ -236,22 +236,17 @@ public string OriginalType
...
             {
                 result = project.ProjectFileExtension;
             }
-            else if (Original is ProjectEvaluation evaluation)
+            else if (Original is ProjectEvaluation evaluation && !string.IsNullOrEmpty(evaluation.ProjectFileExtension))
             {
                 result = evaluation.ProjectFileExtension;
             }
 
-            if (result != null && result != ".sln" && result != ".csproj")
```
yuehuang010, line 250: This original code seems odd to single out csproj.  Is there a reason?

KirillOsenkov: no, it just forgets other project formats. I can't remember if we have a helper that stores all known project file extensions.

KirillOsenkov (approved):

PR #656, opened 2023-02-23
Improve Tracing load times

yuehuang010:
## Switch to OnRender(DrawContext) for Canvas
- Speed up rendering up to 10x faster.
- Simplify TextBlock to reduce creating UI elements.
- Have to redo on_clicks math since TextBlock aren't part of the Canvas Hierarchy.
- Lost mouse over Tooltips ( 😞 ).  Workaround is to double click.
- Load full graph instead of culled subset.  The load times are now fast enough!

## Parallelize Timeline Population (up to 2-3x faster)
- VisitAllChildren itself takes a long time.  There are just a lot of items to go though.
- Move CppAnalysis into one VisitAllChildren, one less loop.

## Misc
- Use case insensitive when parsing $time, $starttime, $endtime
- Search results would show adornment instead of only Framework.
- Fix an issue when mouse zoom would over zoom.

### DIFF
--- a/src/StructuredLogViewer.Core/Timeline/Timeline.cs
+++ b/src/StructuredLogViewer.Core/Timeline/Timeline.cs
@@ -1,13 +1,15 @@
 ﻿using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading;
 using Microsoft.Build.Logging.StructuredLogger;
 
 namespace StructuredLogViewer
 {
     public class Timeline
     {
-        public Dictionary<int, Lane> Lanes { get; set; } = new();
+        public ConcurrentDictionary<int, Lane> Lanes { get; set; } = new();
 
         public Timeline(Build build, bool analyzeCpp)
         {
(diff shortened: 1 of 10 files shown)


src/StructuredLogViewer.Core/Timeline/Timeline.cs
```diff
@@ -62,8 +40,38 @@ private void Populate(Build build, bool analyzeCpp = false)
...
+                     string.Equals(task.Name, "CallTarget", StringComparison.OrdinalIgnoreCase)))
+                {
+                    return;
+                }
+
+                {
+                    var nodeId = timedNode.NodeId;
+                    if (!Lanes.TryGetValue(nodeId, out var lane))
+                    {
+                        lane = new Lane();
```
KirillOsenkov, line 68: I think there's a race condition here where two threads can both create a lane for the same node id

src/StructuredLogViewer/Controls/BuildControl.xaml.cs
```diff
@@ -406,7 +406,10 @@ private void PopulateTrace()
         {
             if (this.tracing.Timeline == null)
             {
+                var start = DateTime.Now;
```
KirillOsenkov, line 409: do UtcNow maybe

PR #664, opened 2023-03-17
Add Nodes Management node for node reusal messages

JanKrivanek:
## Context

Introducing 'Nodes Management' node groupping the new 'Reusing node ...' messages.

## Sample

image


KirillOsenkov:
Thanks!

PR #668, opened 2023-03-27
Add "Count" to "Top 10 Most Expansive Tasks."

yuehuang010:
### Add "Count" to "Top 10 Most Expansive Tasks."
image

### Misc bug fixes
- Fix an issue where Global properties were added to the Properties folder.
- Fix an issue where Item and Properties folders created were when empty.
- Fix an issue where Configuration and Platform adornment were missing.  (Bad Boolean logic)


drewnoakes:
Does that show the number of calls? Perhaps the label could be `123ms, 4 calls`.

KirillOsenkov (approved):

PR #682, opened 2023-05-20
Bring Avalonia version closer to main app.

kant2002:
I also did some formatting changes, so diff would point on the difference more cleanly.


KirillOsenkov (approved):
❤️

Issue #684, opened 2023-05-25
Severe lock contention in BlockingCollection in BinLogReader


rolfbjarne:
The BlockingCollection in BinLogReader suffers from severe lock contention:

https://github.com/KirillOsenkov/MSBuildStructuredLog/blob/786b9739f97b66428a2dde45346d99c47098ad88/src/StructuredLogger/BinaryLogger/BinLogReader.cs#L71-L74

I added some [debug code](https://gist.github.com/rolfbjarne/87a09825dffe9c415183582c74934cd5), and the initial results were:

> Read 24924 records in 00:00:06.8023352 = 6802.3352 ms

raising the bounded capacity from 5000 to 50000, I get:

> Read 24924 records in 00:00:00.3312354 = 331.2354 ms

which means that >95% of the time spent loading the binlog is spent waiting for a lock.

This is on macOS with the Avalonia version, but I'm guessing something similar would happen elsewhere as well.

yuehuang010:
This a buffer to smooth out consumer throughput.  The consumer (constructor) is most likely a bottlenecked on a very expansive event.  Would like to know which event and speed up or moved to a background thread.

KirillOsenkov:
My understanding is that the consumer is now parallel, and it's waiting on the producer.

We can try just increasing the bounded capacity, this way the producer will fill a larger buffer and give the consumer more data to burn through faster.

yuehuang010:
Only the ProjectEvents are spun into a seperate thread because those process a lot of properties.  The rest are remain serialized.  My last profile, the ProjectEvents took up over 60% of the CPU time (when running in series).  There still might be other that could be spun out like Tasks or Targets events.

KirillOsenkov:
So in your opinion, who is slowing things down? The producer or the consumer?

yuehuang010:
Both 😆 .  The producer is more hardware limited as exposed by Apple's better hardware.  The consumer is algorithm limited as all the processing in the CPU.  

Increasing the `boundedCapacity` will unblock for now until someone creates a large enough binlog to hit the cap again.

KirillOsenkov:
Rolf, we're not seeing a wall clock time improvement on Windows after bumping the buffer from 5000 to 50000. I did make the change but will need more thinking about it.

rolfbjarne:
Would just removing the boundedCapacity be a bad idea?

KirillOsenkov:
Don’t know until we measure!

[...]

PR #685, opened 2023-06-02
Add "Buffer" progress bar

yuehuang010:
## Add a "Buffer" progress bar that will report the BlockingCollection queue length.
image
 - The progress bar title is a placeholder, please help suggest a good title.

## Add a "Log Processing Statistics" to report the performance of the each type events.
image

## Add a menu option to clear recently opened history.
image

### DIFF
--- a/src/StructuredLogger/BinaryLog.cs
+++ b/src/StructuredLogger/BinaryLog.cs
@@ -27,7 +27,7 @@ namespace Microsoft.Build.Logging.StructuredLogger
 
         public static Build ReadBuild(string filePath) => ReadBuild(filePath, progress: null);
 
-        public static Build ReadBuild(string filePath, Progress progress)
+        public static Build ReadBuild(string filePath, Progress progress, Progress bufferUsage = null)
         {
             using (var stream = new FileStream(filePath, FileMode.Open, FileAccess.Read, FileShare.Read))
             {
@@ -38,7 +38,7 @@ namespace Microsoft.Build.Logging.StructuredLogger
                     projectImportsArchive = File.ReadAllBytes(projectImportsZipFile);
                 }
 
-                var build = ReadBuild(stream, progress, projectImportsArchive);
+                var build = ReadBuild(stream, progress, bufferUsage, projectImportsArchive);
                 build.LogFilePath = filePath;
                 return build;
             }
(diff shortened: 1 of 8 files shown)


yuehuang010:
#684 FYI.

KirillOsenkov (approved):

src/StructuredLogger/BinaryLog.cs
```diff
@@ -27,7 +27,7 @@ public static IEnumerable<Record> ReadRecords(byte[] binlogBytes)
 
         public static Build ReadBuild(string filePath) => ReadBuild(filePath, progress: null);
 
-        public static Build ReadBuild(string filePath, Progress progress)
+        public static Build ReadBuild(string filePath, Progress progress, Progress bufferUsage = null)
```
KirillOsenkov, line 30: Oof, I hadn't realized that this is breaking public API.

Unfortunately I will have to back it out. Sorry about that.

KirillOsenkov:
I merged this PR optimistically, but upon reflection I think I'm going to have to revert it.

1. I'd like to keep the recent items clearing functionality separate from the rest
2. I'd like to keep the buffer capacity tweaks separate from the rest
3. we can't change the public API as there are too many tools using this library for reading binlogs and we can't require all of them to update just for a progress bar
4. our regular users don't need the buffer utilization progress bar nor do they need the Log Processing Statistics
5. I haven't seen a wall clock time improvement for binlog loading, compared with the previous version. So not sure increasing the buffer is observably helping (at least in the windows case)

src/StructuredLogger/BinaryLog.cs
yuehuang010: No problem.  Do revert.

yuehuang010:
- I will spin up another PR for the Recent Clear function.
- I haven't found an improvement with buff capacity either, but that could be a Windows or hardware difference.  Is there a way to add a "developer" mode flag?

KirillOsenkov:
Don't worry about it, I just pushed the revert:

https://github.com/KirillOsenkov/MSBuildStructuredLog/commit/efb0a575d68d4da534b5e3ba391cc2d29b86be9f

KirillOsenkov:
I would like to avoid regular users paying any perf costs just for developer monitoring tooling.

Your buffer utilization progress bar is good, as well as the perf counters, but stuff like this is good to keep in the branch perhaps. Feel free to make a branch with the revert of https://github.com/KirillOsenkov/MSBuildStructuredLog/commit/efb0a575d68d4da534b5e3ba391cc2d29b86be9f and use it for local testing.

Don't think we need this in the main branch.

src/StructuredLogger/BinaryLog.cs
paulomorgado: Good old overloads are still a thing.

PR #697, opened 2023-07-25
Support corrupt logs produced by some 17.7 MSBuilds

rainersigwald:
Work around bad binlogs produced by https://github.com/dotnet/msbuild/pull/8971 by probing for the zip header that will be the known first bytes of the blob.


rainersigwald:
@michalpavlik can you give this a once-over to see if it matches what we were talking about offline?

KirillOsenkov:
Have we tested this with the corrupt binlogs and the regular binlogs?

rainersigwald:
Yes, it Works On My Machine™️. I've tried with binlogs from current dogfood VS builds, binlogs from the PR with the fix, and binlogs from production 17.6 builds, and verified that they all had many .targets files in the "files" tab.

KirillOsenkov (approved):

PR #700, opened 2023-08-27
WASI support and single-threading

lambdageek:
Cherry pick changes to the core log viewer from the VS Code extension
 https://github.com/lambdageek/MSBuildStructuredLog/tree/wasi-single-thread 

* Add WASI support in PlatformUtilities - the .NET 8 WASI workload doesn't support threads, does not have `Console` colors and cannot assume that `GetTempFileName` will work (because WASI doesn't necessarily grant unrestricted filesystem access)
* Replace a few parallel algorithms by single-threaded versions when threading isn't available.


lambdageek:
I'm not upstreaming the VS Code extension's .NET "engine" or the associated TypeScript code yet, but the changes to the core of the viewer I think are more or less done and I don't expect to need to make many more.

KirillOsenkov (approved):
Very nice, thanks!

Issue #701, opened 2023-08-28
MD5 causes cryptographic exception if FIPS is enabled on Windows


hymccord:
## Description

Recently moved to a new work machine where FIPS is enabled and got an `System.InvalidOperationException` from the `MD5CryptoServiceProvider` when trying to instantiate via `MD5.Create` in the utility method `GetMD5Hash` ([code](https://github.com/KirillOsenkov/MSBuildStructuredLog/blob/afda0c9e14b1bf5615d131c4b1714652cfa37e8d/src/StructuredLogViewer.Core/Utilities.cs#L42)).

## Repro

- Have a windows machine where FIPS is enforced. [How to enable](https://docs.vmware.com/en/VMware-Horizon-Client-for-Windows/2306/horizon-client-windows-installation/GUID-D4238105-E533-4518-A521-B82B49D86585.html)
- Open a binlog
- Right click a project node > Preprocess
- Exception is thrown and you can't view the XML file

## Resolution

I'm more than willing to fix this as it's a very simple fix. Documentation here (https://github.com/microsoft/dotnet/blob/b5f52bf00618f7ccd62b04a5a2fa94b71241a88a/Documentation/compatibility/cryptographicexception-not-thrown-in-fips-mode.md) states that .NetFramework 4.7.2 will throw by default but this was relaxed in .NetFramework 4.8

Either,  
1. Upgrade to .NetFramework to 4.8 so the exception isn't thrown anymore
2. Add `<AppContextSwitchOverrides value="Switch.System.Security.Cryptography.UseLegacyFipsThrow=false" />` to the `runtime` section of `App.config` if planning to stay on .NetFramework 4.7.2

KirillOsenkov:
Thanks, would you mind sending a PR with the app.config change?

hymccord:
Thanks for the quick reply. Sent PR. Tested to make sure it works as well.

PR #702, opened 2023-08-28
Disable FIPS cryptographic exceptions

hymccord:
In .NET Framework 4.7.2 and earlier versions, managed cryptographic provider classes such as System.Security.Cryptography.MD5 throw a System.Security.Cryptography.CryptographicException when the system cryptographic libraries are configured in FIPS mode.

The issue is encountered in the log viewer when needing to write content to a temporary file path e.g. when preprocessing a project node to view the full project XML.

The fix taken here to to turn off the exception throwing behavior via an app runtime switch.

Closes: #701


KirillOsenkov (approved):
:+1:

PR #703, opened 2023-09-05
Support for Extended Data

rokonec:
## Context
MSBuild introduced new concept of Event args Extended data which are supposed to address need to extend events.
One of its most appealing use cases is for passing extended compiler error and warning structured information into msbuild logging infrastructure. Having such info in binlog and viewer might be of use.

### Changes Made
- add support for extended data which will be introduced in FileFormatVersion 17
- derived classes from Error, Warning, and Message including extended data
- add context menu item which prints context of extended data to text window


rokonec:
Attached is binary log for testing if needed.
[binlog.zip](https://github.com/KirillOsenkov/MSBuildStructuredLog/files/12524744/binlog.zip)

KirillOsenkov:
this is nice! I think if extended data is available, we should manufacture a message as a child node of the extended event, and put the full text in there. Also if extended metadata is present, add metadata child nodes for each metadata

KirillOsenkov:
See `Construction.AddProperties` for an example, but maybe use Metadata instead of Property

src/StructuredLogViewer/Controls/BuildControl.xaml.cs
```diff
@@ -56,6 +56,7 @@ public partial class BuildControl : UserControl
         private MenuItem debugItem;
         private MenuItem hideItem;
         private MenuItem showTimeItem;
+        private MenuItem showExtendedData;
```
KirillOsenkov, line 59: don't forget to remove this

src/StructuredLogger/Construction/Construction.cs
```diff
@@ -780,8 +782,57 @@ private void Populate(AbstractDiagnostic message, BuildErrorEventArgs args)
             message.File = Intern(args.File);
             message.ProjectFile = Intern(args.ProjectFile);
             message.Subcategory = Intern(args.Subcategory);
+
+            PopulateWithExtendedData(message, args);
         }
 
+        internal static void PopulateWithExtendedData(TreeNode node, BuildEventArgs args)
+        {
+            if (args is IExtendedBuildEventArgs extended)
```
KirillOsenkov, line 791: I'd flip this condition:

```
if (args is not IExtendedBuildEventArgs extended)
{
    return;
}
```

src/StructuredLogViewer/themes/Generic.xaml
```diff
@@ -549,7 +550,7 @@
         <Setter TargetName="messageText" Property="Opacity" Value="0.25" />
       </DataTrigger>
     </DataTemplate.Triggers>
```
KirillOsenkov, line 552: Probably need to change these to HierarchicalDataTemplate.Triggers

KirillOsenkov (approved):

src/StructuredLogViewer/themes/Generic.xaml
KirillOsenkov: Surprised that it works, but OK

PR #711, opened 2023-10-27
Sensitive data redacting PoC

JanKrivanek:
### Context
Based on offline talks with @KirillOsenkov I'm spinning some experimentation with sensitive data redacting.
Currently the main effort was on:
 * bringing the Write and Replay functionality back to functional state
 * adding ability to replay embedded content
 * adding ability to hook embedded strings and embedded files

The actual redacting is a very PoC thing - just to demonstrate the idea.
In next iteration I'd ideally ref the https://www.nuget.org/packages/MSBuild.BinlogRedactor/ and just inject the replay functionality into it. The package would take care about the redaction of data - as it already has some auto-detection of sensitive data (and more patterns are going to be added soon)

If all the stream utilities in `Postprocessing` folder feels as too much code to drag - I can reference https://www.nuget.org/packages/DotUtils.StreamUtils/,  but viewer seeems to be very minimalistic in package refs - so I rather just used the code.

### UX
Updated:

image
<br/>
image
<br/>
image
<br/>
image


JanKrivanek:
The CI is failing due the new test getting
```
System.IO.FileNotFoundException : Could not find file 'C:\Users\appveyor\AppData\Local\Temp\1\47b8ec7b-5245-49cf-8842-ef31ceaa4965\47b8ec7b-5245-49cf-8842-ef31ceaa4965\assembly\dl3\39c84884\6d30ae17_e408da01\1.binlog'.
```

While for obtaining that test binlog (1.binlog) I use same `GetTestFile("1.binlog")` call: https://github.com/KirillOsenkov/MSBuildStructuredLog/pull/711/files#diff-1ab0219f9bea23d8aa114270d3b3010a73c59dbca476befba614f2e9cb6f89d3R132

as is already used in other tests - e.g.:

https://github.com/KirillOsenkov/MSBuildStructuredLog/blob/ff5383d7849450a80306eda846b07d419c63c6f7/src/StructuredLogger.Tests/BinaryLoggerTests.cs#L93-L95

This works just fine locally.
@KirillOsenkov - any quick idea from top of your head, why `1.binlog` might be missing?

KirillOsenkov:
No quick idea, just noticing that in the path the GUID is mentioned twice:
47b8ec7b-5245-49cf-8842-ef31ceaa4965\47b8ec7b-5245-49cf-8842-ef31ceaa4965

probably eyeball `GetTestFile()` and see where it could go wrong?

KirillOsenkov:
Some first notes as I read:

 1. better call it "secrets" everywhere, not "passwords"
 2. avoid abbreviations, so "password" instead of "pwd" for example
 3. I'd avoid using `Path.GetTempFileName()`, it can only hold 65,536 files, can overflow easily (and it's hard to debug), and people don't clean it up. Use the viewer directory (`%localappdata%\Microsoft\MSBuildStructuredLog\Temp\`), I think we already have code somewhere to write into it.
 4. I think there was a problem with making BuildEventArgsReader public, but I don't quite remember what it is. It could be that it would conflict with the one in MSBuild?
 5. when you update to latest main you'll notice Roman's recent change to support ExtendedCriticalBuildMessageEventArgs
 6. in terms of UI, you could probably show an empty editor in the right pane (there's an easy API for that), and then use each line from that text as a secret, and have a button to "commit". This way you could maybe specify an existing file with secrets?
 7. A design decision that needs to be made is whether to redact in-place or into a different file.

Otherwise all seems reasonable from a quick read.

JanKrivanek:
> Some first notes as I read:
> 
> 1. better call it "secrets" everywhere, not "passwords"
> 2. avoid abbreviations, so "password" instead of "pwd" for example
> 3. I'd avoid using `Path.GetTempFileName()`, it can only hold 65,536 files, can overflow easily (and it's hard to debug), and people don't clean it up. Use the viewer directory (`%localappdata%\Microsoft\MSBuildStructuredLog\Temp\`), I think we already have code somewhere to write into it.
> 4. I think there was a problem with making BuildEventArgsReader public, but I don't quite remember what it is. It could be that it would conflict with the one in MSBuild?
> 5. when you update to latest main you'll notice Roman's recent change to support ExtendedCriticalBuildMessageEventArgs
> 6. in terms of UI, you could probably show an empty editor in the right pane (there's an easy API for that), and then use each line from that text as a secret, and have a button to "commit". This way you could maybe specify an existing file with secrets?
> 7. A design decision that needs to be made is whether to redact in-place or into a different file.
> 8. Are you sure we need several new interfaces here, all fine-granular? It looks good, but just check that every interface is justified, or can we make do without?
> 
> Otherwise all seems reasonable from a quick read.

I'l work on these.

As far as UI - if you have anything specific in mind, a quick non-functional markup mockup would be super helpful as I'm bit rusty with UI. But I can definitely find a way otherwise. I'd probably add the 'Replace In Place' checkbox (or something similar that would flip behavior similar to Save / Save As) - that way we wouldn't need to decide whether it's a in-place redact or separate file - it'd be up to user, with sensible default.

Interfaces  - I'd want to keep this as similar to MSBuild as possible - it'll make it easier to port changes both ways. But I'm definitely open to suggetions here

JanKrivanek:
I hope I addressed all the preliminary comments + I created a proposal for the GUI.


So the current GUI proposal:

image
<br/>
image
<br/>
image

 * The choice to overwrite in place vs save as new is left upon the user
 * Redacting is using the BuildProgress control to prevent interaction during redacting
 * Few other choices are exposed - the 'common credentials' option is currently disabled (with 'not yet supported' tooltip). But I'll plan to pull and reference the 'real' redactor nuget in next revision and hence got the additional redactors functionality

JanKrivanek:
Finalized v1:
* Removed temporary code, replaced with package references (redactor and stream utils) - this as well brings the automated detection of some common secret types (more are going to be added to the package soon).
* Surfaced the additional redacting options in UI (updated the GUI screens in the description of this PR)

It feels ready for review ;-)

src/StructuredLogger/BinaryLogger/BuildEventArgsReader.cs
```diff
@@ -250,23 +352,22 @@ private void ReadBlob(BinaryLogRecordKind kind)
...
             else
             {
                 int length = ReadInt32();
-                byte[] bytes = binaryReader.ReadBytes(length);
-                OnBlobRead?.Invoke(kind, bytes);
+                return binaryReader.BaseStream.Slice(length);
             }
         }
 
+        private void SkipBytes(int count)
```
KirillOsenkov, line 366: Is this used?

src/StructuredLogger/BinaryLogger/Postprocessing/BinlogRedactor.cs
```diff
@@ -0,0 +1,146 @@
+﻿using System;
+using System.Linq;
+using System.Collections.Generic;
+using System.Text;
+using Microsoft.Build.Logging;
+using Microsoft.Build.Logging.StructuredLogger;
+using System.IO;
+using System.Threading;
+using Microsoft.Build.SensitiveDataDetector;
```
KirillOsenkov, line 9: Sort usings

src/StructuredLogger/BinaryLogger/Postprocessing/StringReadEventArgs.cs
```diff
@@ -0,0 +1,35 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+
+namespace Microsoft.Build.Logging
+{
+    /// <summary>
+    /// An event args for <see cref="IBuildEventStringsReader.StringReadDone"/> callback.
+    /// </summary>
+    public sealed class StringReadEventArgs : EventArgs
```
KirillOsenkov, line 11: One of my more serious concerns is that this is a class, and we'll allocate it for every string (but only if someone is listening). The binlog I'm currently looking at has 2.2 million strings, at 40 bytes per object on 64-bit it'll be around 100 MB allocations right there.

Wondering if we can find a way to avoid allocating per string here, but if it's hard, it's OK. As long as we make sure the normal path of opening binlogs is not impacted negatively.

I've worked extremely hard to make sure binlog opening is as fast as possible.

src/StructuredLogger/StructuredLogger.csproj
```diff
@@ -28,6 +28,8 @@
     <PackageReference Include="Microsoft.Build.Framework" Version="$(MSBuildPackageVersion)" />
     <PackageReference Include="Microsoft.Build.Utilities.Core" Version="$(MSBuildPackageVersion)" />
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="all" />
+    <PackageReference Include="MSBuild.BinlogRedactor.SensitiveDataDetector" Version="0.0.4-beta" />
+    <PackageReference Include="DotUtils.StreamUtils" Version="0.0.6" />
```
KirillOsenkov, line 32: I'm hesitant about adding dependencies to the core StructuredLogger.dll and the MSBuild.StructuredLogger NuGet package. I think StreamUtils is only used for a few helpers, such as `ReadToEnd`, `SkipBytes`, `Concat` and `Slice`. It is of course neat to be able to have primitives like these to operate on streams, I'm just wondering if it'd be possible to inline this logic into a .cs file and include that into the project to avoid a dependency. 

Otherwise we'd have to find a way to ship a non-prelease NuGet package depending on a couple of pre-release ones.

src/StructuredLogger/StructuredLogger.csproj
```diff
@@ -28,6 +28,8 @@
     <PackageReference Include="Microsoft.Build.Framework" Version="$(MSBuildPackageVersion)" />
     <PackageReference Include="Microsoft.Build.Utilities.Core" Version="$(MSBuildPackageVersion)" />
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="all" />
+    <PackageReference Include="MSBuild.BinlogRedactor.SensitiveDataDetector" Version="0.0.4-beta" />
```
KirillOsenkov, line 31: Same concern with adding another dependency here. Do you think it'd be possible to only have the `StructuredLogViewer` and `BinlogTool` projects depend on this one? So that people who only depend on StructuredLogger.dll do not inherit two more dependencies they probably don't need? This would mean that StructuredLogger.dll doesn't have the built-in ability to redact.

src/StructuredLogger/StructuredLogger.cs
```diff
@@ -103,10 +105,14 @@ public override void Shutdown()
 
             if (projectImportsCollector != null)
             {
-                var bytes = projectImportsCollector.GetAllBytes();
-                construction.Build.SourceFilesArchive = bytes;
-
                 projectImportsCollector.Close();
+
+
```
KirillOsenkov, line 110: I like to avoid two consecutive empty lines in .cs files, perhaps I should add an .editorconfig rule for this.

src/StructuredLogger/BinaryLogger/Postprocessing/BinlogRedactor.cs
```diff
@@ -0,0 +1,146 @@
...
+using System.Linq;
+using System.Collections.Generic;
+using System.Text;
+using Microsoft.Build.Logging;
+using Microsoft.Build.Logging.StructuredLogger;
+using System.IO;
+using System.Threading;
+using Microsoft.Build.SensitiveDataDetector;
+
+namespace StructuredLogger.BinaryLogger.Postprocessing
```
KirillOsenkov, line 11: I have to confess that the namespace naming is historically kind of a mess in the MSBuildStructuredLog repo. I'm hoping to move towards rationalizing it all a bit.

For this one, how about `Microsoft.Build.Logging.BinaryLogger.Postprocessing`? Or maybe just leave `Microsoft.Build.Logging.BinaryLogger`?

Also, would it be possible to move this type out to StructuredLogViewer.Core perhaps? If we're going to try to avoid having StructuredLogger.dll to have a dependency on the redaction logic.

src/StructuredLogger/BinaryLogger/Postprocessing/ArchiveFileEventArgs.cs
```diff
@@ -0,0 +1,19 @@
...
+namespace Microsoft.Build.Logging;
+
+/// <summary>
+/// Event arguments for <see cref="IBuildFileReader.ArchiveFileEncountered"/> event.
+/// </summary>
+public sealed class ArchiveFileEventArgs : EventArgs
+{
+    public ArchiveFileEventArgs(ArchiveFile archiveFile) =>
+        ArchiveFile = archiveFile;
+
```
KirillOsenkov, line 16: no two consecutive empty lines in .cs files

src/StructuredLogger/Progress.cs
```diff
@@ -9,6 +9,11 @@ public class Progress : IProgress<ProgressUpdate>
 
         public event Action<ProgressUpdate> Updated;
 
+        public virtual void Report(double ratio)
```
KirillOsenkov, line 12: Oops, I just removed it. Do you need it or could you use the ProgressUpdate?

src/StructuredLogViewer/MainWindow.xaml.cs
```diff
@@ -356,7 +358,7 @@ private void UpdateRecentItemsMenu(WelcomeScreen welcomeScreen = null)
         private void SetContent(object content)
         {
             // We save build control to allow to bring back states to new one
-            if (mainContent.Content is BuildControl current)
+             if (mainContent.Content is BuildControl current)
```
KirillOsenkov, line 361: what is this :)

src/StructuredLogViewer/MainWindow.xaml.cs
```diff
@@ -659,31 +663,119 @@ private void Reload()
             OpenLogFile(logFilePath);
         }
 
-        private void SaveAs()
+        private async void RedactSecrets()
```
KirillOsenkov, line 666: async void is sketchy for various reasons :)

I usually use a pattern like this:

```
void Foo()
{
    FooAsync().Ignore(); // Ignore 
    _ = FooAsync(); // or like this, "observes" the task
}

async Task FooAsync()
{
}

        /// <summary>
        /// Consumes a task and doesn't do anything with it.  Useful for fire-and-forget calls to async methods within sync methods.
        /// </summary>
        /// <param name="task">The task whose result is to be ignored.</param>
        public static void Ignore(this Task task)
        {
        }
```

src/StructuredLogger/BinaryLogger/BinaryLogReplayEventSource.cs
```diff
@@ -0,0 +1,182 @@
...
+        IBuildFileReader,
+        IEmbeddedContentSource
+    { }
+
+    /// <summary>
+    /// Provides a method to read a binary log file (*.binlog) and replay all stored BuildEventArgs
+    /// by implementing IEventSource and raising corresponding events.
+    /// </summary>
+    /// <remarks>The class is public so that we can call it from MSBuild.exe when replaying a log file.</remarks>
+    public sealed class BinaryLogReplayEventSource : EventArgsDispatcher, IBinaryLogReplaySource
```
KirillOsenkov, line 28: What's your thinking for this type? Is this similar to the one in MSBuild? What should this type be used for? Only in the viewer repo?

src/StructuredLogger/BinaryLogger/BuildEventArgsReader.cs
```diff
@@ -203,10 +224,91 @@ private static bool IsAuxiliaryRecord(BinaryLogRecordKind recordKind)
         private void ReadBlob(BinaryLogRecordKind kind)
```
KirillOsenkov, line 224: In this entire ReadBlob() method we should be very mindful of allocations, and be explicit about every allocation happening. I'm dealing with a binlog that is 900 MB in size, 600 MB of which are the embedded .nuget.dgspec.json files (https://github.com/NuGet/Home/issues/12957). They expand to over 10 GB in memory.

I haven't read this closely, but let's make sure every allocation is accounted for. Ideally I also wanted to make ArchiveFiles lazy if possible, so that they decompress their file content only when requested, to save on memory. But that's for later. For now let's just double-check that this only allocates where necessary and avoids unnecessary allocations.

src/StructuredLogger/BinaryLogger/BuildEventArgsReader.cs
```diff
@@ -203,10 +224,91 @@ private static bool IsAuxiliaryRecord(BinaryLogRecordKind recordKind)
...
+            }
+            else
+            {
+                embeddedStream.SkipBytes();
+            }
+        }
+
+        private void InvokeEmbeddedDataListeners(BinaryLogRecordKind kind, Stream embeddedStream)
+        {
+            byte[] bytes = null;
```
KirillOsenkov, line 286: Let's see who currently needs the giant byte array, perhaps we can get rid of it. It's a huge source of memory usage. Ideally I think an IReadOnlyList<ArchiveFile> is all the consumers should need.

I've been planning to optimize how we store the byte[] array and process it, but I decided to wait for your PR to land first to avoid causing a messy merge.

src/StructuredLogger/BinaryLogger/BuildEventArgsReader.cs
```diff
@@ -250,23 +352,22 @@ private void ReadBlob(BinaryLogRecordKind kind)
...
-
-                byte[] fullBytes = new byte[length];
-                prefixBytes.CopyTo(fullBytes, 0);
-                bytes.CopyTo(fullBytes, prefixBytes.Length);
-                bytes = fullBytes;
-
-                OnBlobRead?.Invoke(kind, bytes);
+                Stream prefixStream = new MemoryStream(prefixBytes);
+                Stream dataStream = binaryReader.BaseStream.Slice(length - prefixBytes.Length);
+                return prefixStream.Concat(dataStream);
```
KirillOsenkov, line 357: This is neat and if implemented right, also much cheaper than byte arrays. Just worried about an extra dependency.

I also wasn't happy when we added Microsoft.NET.StringTools.dll as a core dependency to MSBuild when I was not looking.

src/StructuredLogger/BinaryLogger/BuildEventArgsReader.cs
```diff
@@ -1448,9 +1549,17 @@ private IEnumerable ReadTaskItemList()
             return list;
         }
 
+        private readonly StringReadEventArgs stringReadEventArgs = new StringReadEventArgs(string.Empty);
         private string ReadString()
         {
-            return binaryReader.ReadString();
+            string text = binaryReader.ReadString();
+            if (this.StringReadDone != null)
```
KirillOsenkov, line 1556: I think if nobody subscribes to this, then we won't allocate, which seems fine.

src/StructuredLogger/BinaryLogger/BuildEventArgsWriter.cs
```diff
@@ -250,10 +290,17 @@ private void Write(BuildStartedEventArgs e)
...
             else
             {
-                Write(0);
+                Write(e.BuildEnvironment?.Where(kvp => IsWellKnownEnvironmentDerivedProperty(kvp.Key)));
             }
         }
 
+        private static bool IsWellKnownEnvironmentDerivedProperty(string propertyName)
+        {
+            return propertyName.StartsWith("MSBUILD") ||
```
KirillOsenkov, line 299: Should these use StringComparison.OrdinalIgnoreCase? Or nah?

src/StructuredLogger/BinaryLogger/BuildEventArgsWriter.cs
```diff
@@ -422,6 +471,7 @@ private void Write(BuildMessageEventArgs e)
         {
             switch (e)
             {
+                case FileUsedEventArgs responseFileUsed: Write(responseFileUsed); break;
```
KirillOsenkov, line 474: why did FileUsed move to the top? I'm not sure if the order here matters and how the switch statement works for types under the covers. Generally I'd put most common event args on top, and FileUsed is probably the least commonly used of them all.

src/StructuredLogger/BinaryLogger/BuildEventArgsWriter.cs
```diff
@@ -456,12 +505,24 @@ private void Write(TargetSkippedEventArgs e)
             WriteDeduplicatedString(e.TargetFile);
             WriteDeduplicatedString(e.TargetName);
             WriteDeduplicatedString(e.ParentTarget);
-            WriteDeduplicatedString(""); // Condition
-            WriteDeduplicatedString(""); // EvaluatedCondition
-            Write(true); // OriginallySucceeded
+            WriteDeduplicatedString(e.Condition); // Condition
```
KirillOsenkov, line 508: wow, looks like our copy of the Writer was indeed a bit behind! I sure hope nobody was using it for anything important!

src/StructuredLogger/BinaryLogger/BuildEventArgsReader.cs
```diff
@@ -203,10 +224,91 @@ private static bool IsAuxiliaryRecord(BinaryLogRecordKind recordKind)
...
+            {
+                // We could create ZipArchive over the target stream, and write to that directly,
+                //  however, binlog format needs to know stream size upfront - which is unknown,
+                //  so we would need to write the size after - and that would require the target stream to be seekable (which it's not)
+                ProjectImportsCollector? projectImportsCollector = null;
+
+                if (EmbeddedContentRead != null)
+                {
+                    projectImportsCollector =
+                        new ProjectImportsCollector(Path.GetRandomFileName(), false, runOnBackground: false);
```
KirillOsenkov, line 240: Wait, didn't we have a helper for random temp files?

KirillOsenkov:
Thank you for this gargantuan effort! Your hard work and patience are much appreciated! I think it looks really good already.

I really didn't want to cause any unnecessary stop energy, but I'm paranoid about breaking people, so just wanted to iron out a couple of things before this goes in.

Primarily:
 1. do you think it's doable to move the dependency on redaction .dll and the redactor logic from the StructuredLogger.dll and into the "UI" and BinlogTool layer? Technically I'd like to keep StructuredLogger minimal, just read/write binlogs, no UI, no extra stuff. People should be able to build on top of that to add extra functionality. Perhaps StructuredLogViewer.Core is a better place? It's "in between" the core and the UI layers. Or maybe add a stub empty static Redact() delegate in the core, and have the UI plug the real implementation.
 2. similarly, if we could avoid the dependency on the streams .dll... I know it's ugly to inline the source, but historically people have been requesting that it's possible to just download StructuredLogger.dll from the Release and be able to use it, even bypassing the NuGet package entirely. This is why I've been publishing StructuredLogger.dll as part of every release: https://github.com/KirillOsenkov/MSBuildStructuredLog/releases/tag/v2.1.858
 3. is the redaction workflow separate from the regular binlog reading logic? If so, I have no concerns. I'm just worried to not regress the mainline binlog loading scenarios in terms of performance.

The rest (including the UI) seems fine. Once the above is clarified, we could merge and continue iterating on it in separate PRs if needed.

For the binlogtool, we should also later think about supporting a ".rsp" file, so that someone has a file with secrets somewhere (one per line), and we could just pass it to the tool instead of mentioning the secrets in the command line. Also clarify the in-place vs. a new file for command-line.

src/StructuredLogViewer/MainWindow.xaml.cs
```diff
@@ -659,31 +663,119 @@ private void Reload()
             OpenLogFile(logFilePath);
         }
 
-        private void SaveAs()
+        private async void RedactSecrets()
         {
-            if (currentBuild != null)
+            RedactInputControl redactInputControl = new RedactInputControl(GetSaveAsDestination);
+            if (redactInputControl.ShowDialog() == true)
```
KirillOsenkov, line 669: Let's invert this if, and if the result is not true, return early. This would allow to un-indent the rest of the method.

src/StructuredLogViewer/MainWindow.xaml.cs
```diff
@@ -659,31 +663,119 @@ private void Reload()
...
+                {
+                    // Reload
+                    OpenLogFile(logFilePath);
+                }
+                else
+                {
+                    SetContent(currentBuild);
+                    AnnounceFileSaved(redactInputControl.DestinationFile);
+                }
+
```
KirillOsenkov, line 739: no need for an empty line between two closing curlies. I should codify this in .editorconfig too.

src/StructuredLogger.Tests/BinaryLoggerTests.cs
```diff
@@ -124,6 +124,90 @@ public void TestBinaryLoggerRoundtrip(bool useInMemoryProject)
...
+            Dictionary<string, string> embedFiles2 = new();
+
+            reader1.ArchiveFileEncountered += arg
+                => AddArchiveFile(embedFiles1, arg);
+            reader2.ArchiveFileEncountered += arg
+               => AddArchiveFile(embedFiles2, arg);
+
+
+            int i = 0;
+            while (reader1.Read() is { } ev1)
```
KirillOsenkov, line 190: let's use a longer variable name that makes it clear what it is

src/StructuredLogger.Tests/StructuredLogger.Tests.csproj
```diff
@@ -32,4 +33,9 @@
   <ItemGroup>
     <Service Include="{82a7f48d-3b50-4b1e-b82e-3ada8210c358}" />
   </ItemGroup>
+  <PropertyGroup Condition=" '$(OS)' == 'Windows_NT' ">
+    <SignAssembly>True</SignAssembly>
+    <AssemblyOriginatorKeyFile>..\StructuredLogger\key.snk</AssemblyOriginatorKeyFile>
+    <DelaySign>False</DelaySign>
```
KirillOsenkov, line 39: I'm guessing this is for IVT. Which API is internal that you'd like to use from tests?

src/StructuredLogger/BinaryLogger/BinaryLogReplayEventSource.cs
KirillOsenkov: Oh, I guess this is temporary until we update to the shipped version of the MSBuild NuGet that has this? Is that right?

KirillOsenkov:
Hmm, I just thought, for Stream utils, do you think you could publish a source-only package? DotUtils.StreamUtils.Source? You could build it from the same source using for example a custom .nuspec file: https://github.com/KirillOsenkov/MSBuildTools/blob/d8afc187fa7be3771743e223a31765a68248f4e5/src/PackNuspec/PackNuspec.csproj#L9-L37 

or coerce the built-in MSBuild properties to produce a source-only package in addition to the default one. 

I think this would be ideal: doesn't add a runtime dependency, but also avoids forking the source. Wish we could do the same for Microsoft.NET.StringTools.dll, but that ship has sailed I'm afraid.

src/StructuredLogger/BinaryLogger/Postprocessing/StringReadEventArgs.cs
JanKrivanek: Right. I definitely have an appreciation for any perf efforts :-)

In this case there is a single instance of the type per entire log reading and it's being reused between event calls: https://github.com/KirillOsenkov/MSBuildStructuredLog/pull/711/files#diff-5850999507dea96dddb234354cf875f2e8ba46b9f7d0e0dd1019bbf137e3e60bR1558

The situation is explicitly documented on the event declaration:
https://github.com/KirillOsenkov/MSBuildStructuredLog/pull/711/files#diff-6f29c5341d4c9e3afce3ea5138039ac5d6aee77639c0744de193a7a7b9c6ed5dR16

```
/// The passed event arg can be reused and should not be stored.
```

KirillOsenkov:
Sounds great!

StructuredLogger only references these:

image

And Microsoft.Build.Utilities.Core.dll is only used for the `Logger` type:
image

I only mentioned Microsoft.NET.StringTools.dll as an example of a dependency MSBuild didn't add (or added as source-only), because it caused a lot of grief for a lot of tools out there, that use MSBuild libraries, they now needed to ship an extra .dll, update installers, etc.

src/StructuredLogger/StructuredLogger.cs
JanKrivanek: I wanted to add a rule - but it seems it's not possible :-( https://github.com/editorconfig/editorconfig/issues/336#issuecomment-372373305

src/StructuredLogger/Progress.cs
JanKrivanek: This method is not necessary but keeps the calling code more simple. Would you rather see it disappear?

src/StructuredLogger/BinaryLogger/BinaryLogReplayEventSource.cs
JanKrivanek: Right. I made it `internal`

src/StructuredLogger/BinaryLogger/BuildEventArgsReader.cs
JanKrivanek: I'm resolving this comment now, as the current way how embedded files are read is unchanged by this PR.

Changing all consumers to accept `Stream` instead would be great. It can get bit tricky - as it needs to be consumed at the time of invoking the event and cannot be shared by multiple consumers - both of which seem to be broken by current consuming patterns (but I've looked just very quickly).

Alternatively some tricks might possibly be used to allow lazy delayed consuming - they would boil down to reading from disk (either remembering position and size and then decompress and skip to the embedded file on demand, or storig the embedded archive into separate temporary file) - this can cause slower load of embedded files (but faster initial load of just events).

In ideal world we'd change the binlog format - the compressed events and then the archive would be each compressed independently and placed in a single file with short 'table of content' on start. That way we can read just the events during the initial load (faster -> win) and then when files would be needed, we can quickly switch to stream position with the archive and read files on demand (memory efficient -> win)

src/StructuredLogger/BinaryLogger/BuildEventArgsWriter.cs
JanKrivanek: Nah :-)
It's same in MSBuild

https://github.com/dotnet/msbuild/blob/5d1509792899a9ac25f59b8b0034dd98989a30b1/src/Shared/EnvironmentUtilities.cs#L16-L21

```csharp
        public static bool IsWellKnownEnvironmentDerivedProperty(string propertyName)
        {
            return propertyName.StartsWith("MSBUILD") ||
                propertyName.StartsWith("COMPLUS_") ||
                propertyName.StartsWith("DOTNET_");
        }
```

src/StructuredLogger/BinaryLogger/BuildEventArgsWriter.cs
JanKrivanek: Makes sense.
Though I'd keep the same ordering as in MSBuild for less painfull diffing.
IIRC it should translate to dictionary - so hopefully no harm

src/StructuredLogger.Tests/StructuredLogger.Tests.csproj
JanKrivanek: Correct.
`BuildEventArgsReader` is needed (to verify the equality of full roundtrip through reader and writer - as that's needed for redacting).

src/StructuredLogger/Progress.cs
KirillOsenkov: No, it’s fine, we can keep it

src/StructuredLogger/StructuredLogger.csproj
JanKrivanek: I added a new indirection project (`StructuredLogger.Utils`) to separate the functionality from the `StructuredLogger`.
I wanted to avoid duplicating or linking the redaction work in both of the ui layers. (The existing `StructuredLogViewer.Core` is not suitable for binlogtool and referencing binlogtool from the viewer felt too unnecesary)

KirillOsenkov: Sounds good. Let's make sure all .dlls that need to ship with the viewer are mentioned here:
https://github.com/KirillOsenkov/MSBuildStructuredLog/blob/main/src/StructuredLogViewer/MSBuildStructuredLogViewer.nuspec

For binlogtool, I think the packing is all automatic, so nothing should be done hopefully.

src/StructuredLogger/BinaryLogger/BuildEventArgsWriter.cs
KirillOsenkov: Sure, if MSBuild does that, we should do the same think. I'm also not sure if the order even matters here.

src/StructuredLogger.Tests/StructuredLogger.Tests.csproj
KirillOsenkov: Oh, and we can't make that public because otherwise it'll conflict with the one in the actual MSBuild for third-parties? Makes sense.

src/StructuredLogger/BinaryLogger/Postprocessing/BinlogRedactor.cs
JanKrivanek: I moved all files in Postprocessing folder to `Microsoft.Build.Logging.StructuredLogger` namespace

src/StructuredLogger/BinaryLogger/BuildEventArgsReader.cs
JanKrivanek: https://github.com/KirillOsenkov/MSBuildStructuredLog/pull/711#discussion_r1384920463 applies here as well :-)

Basically - the existing (reading) workflow should be untouched.
However it still means that during reading the entire embedded archive is greedily decompressed and transformed to files in memory. This can definitely be improved. Ideally with one more binlog format breaking change (but that's for a separate talk)

src/StructuredLogger/BinaryLogger/BuildEventArgsReader.cs
JanKrivanek: Thank you for learning me something new with the source only packages! ;-)

src/StructuredLogger.Tests/StructuredLogger.Tests.csproj
JanKrivanek: Correct.
We can possibly place it in different namespace to avoid the conflict. But not exposing at all feels as a better way.

src/StructuredLogger/StructuredLogger.csproj
JanKrivanek: So for binlogtool I installed it from local built package and verified it works as expected - good.
For the viewer I verified by building, extracting the lib folder from the created packages and verifying it works as expected - not sure if there is better way of testing.

JanKrivanek:
@KirillOsenkov - it now feels ready for next round of review.

Main changes:
 * Addressed all feedback (hopefully)
 * Manually e-2-e tested (not sure if there is recommended way. WhatI did: https://github.com/KirillOsenkov/MSBuildStructuredLog/pull/711#discussion_r1386660979)
 * Added indirection StructuredLogger.Utils for separating redaction logic and dependency from the StructuredLogger
 * StreamUtils added as sources
 * Improved redaction via binlog tool (still not all options exposed, but already usable for scraping logs in CIs)

Looking forward for the next round of feedback :-)

KirillOsenkov (approved):
Awesome job! Thanks so much!

JanKrivanek:
Oh - nice
Please ping me once you'll publish new version of the binlogtool - I'll then work on incorporating into arcade

KirillOsenkov:
Published:

 * https://github.com/KirillOsenkov/MSBuildStructuredLog/releases/tag/v2.2.1
 * https://www.nuget.org/packages/MSBuild.StructuredLogger/2.2.2

KirillOsenkov:
image

image

PR #732, opened 2023-12-08
Add support for forward compatible binlog format

JanKrivanek:
## Context
Complement to MSBuild PR https://github.com/dotnet/msbuild/pull/9307
This introduces capability to understand the new log format that contains event offsets - with those, reader can detect issues with individual events and choose how to handle those (recover/skip/fail). Logs now as well contain the minimum needed version of reader - to capture situations where new log format changes beyond comprehension ability of particular reader version.

User is still given the ability to decide the level of compatibility and see the resulting stats of compatability issues.

### Further info
Some perf stats: https://github.com/JanKrivanek/msbuild/pull/1
MSBuild parent bug: https://github.com/dotnet/msbuild/issues/9261

### GUI

image

### CLI

Unchanged

### API

The `Serialization.Read` method will accept optional `ReaderSettings` with `UnknownDataBehavior` parameter (of an enum type with the same name). If not specified, the compatibility issues are summarized in a single synthetized Error node on the Build.

```csharp
   // This will work just fine on future versions of binlog
   var build = Serialization.Read(logPath);
```

---------------------------
_stop reading from here below - captures historic evolution of the PR that is not part of the final version_

### GUI

image

`Cancel`:

image

`OK`:

image

`File` -> `Stats` (upon expanding all collapsed nodes):

image

### <a name="cli"></a>CLI

image

### <a name="api">API

The `Serialization.Read` method will accept optional `IForwardCompatibilityReadSettings` - if not specified, the legacy behavior (exception on newer format) is preserved:

```csharp
   // This will throw on future version of logs
   var build = Serialization.Read(logPath);
```

To leverage the new behavior - one need to specify the forward compatibility handler - e.g.:

```csharp
   // This will allow processing future version of logs and will store the reading errors summary in build.RecoverableReadingErrors
   var build = Serialization.Read(logPath, progress: null, forwardCompatibilitySettings: ((AllowForwardCompatibilityDelegate)(_ => true)).WithDefaultHandler());
```

The handler allows granular control of opt-ing into the forward compatibility behavior (in case incompatibility is encountered) and handling the individual reading errors. For summary the reading errors are stored in `Build.RecoverableReadingErrors` property.

Alternatively a `ForwardCompatibilityReadingHandler` from binlogtool can be used:

```csharp
  ForwardCompatibilityReadingHandler compatHandler = new ForwardCompatibilityReadingHandler();
  compatHandler.SetMode(ForwardCompatibilityReadingHandler.Mode.LogErrorsDetailed);
  var build = compatHandler.ReadBuild(logPath);
  // optional summarization of errors
  compatHandler.HandleBuildResults(build, Console.Error);
```


KirillOsenkov:
My initial gut reaction is that forward compat should be the default for all readers, and if there are compat issues we can emit a synthetic `BuildErrorEventArgs` or `BuildWarningEventArgs`. Maybe consumers could tweak whether they want to see an error, a warning, a message or nothing at all for compat issues.

So I'd remove things like ForwardCompatibilityExplicitlyConfigured, -fwd, and related things and keep it as simple as possible. Since the current behavior is we throw an exception, literally anything else is an improvement. I'd like users to update to this version of the NuGet and not have to change anything and be automatically opted in. 

The new GUI seems nice, but I'd think it's an overkill (just display a synthetic Error or Warning like I suggest above at the very top of the tree). But since you've already built it, might as well leave it I guess. I think users won't see this often anyway, since the viewer updates when it starts.

src/StructuredLogger/BinaryLogger/Postprocessing/ReaderErrorType.cs
```diff
@@ -0,0 +1,29 @@
...
+
+/// <summary>
+/// Type of the error that occurred during reading.
+/// </summary>
+public enum ReaderErrorType
+{
+    /// <summary>
+    /// The encountered event is completely unknown to the reader. It cannot interpret any part of it.
+    /// </summary>
+    UnkownEventType,
```
KirillOsenkov, line 14: Unknown

src/StructuredLogger/ObjectModel/Build.cs
```diff
@@ -288,5 +289,18 @@ public void WaitForBackgroundTasks()
...
+        /// Indicates whether current log/build is of a higher than known version
+        ///  or that there were unknown events or data detected during reading.
+        /// </summary>
+        public bool IsCompatibilityMode { get; internal set; }
+
+        /// <summary>
+        /// In case forward compatible reading was allowed and encountered recoverable errors, this method
+        ///  will return the list of errors encountered during reading.
+        /// </summary>
+        public IReadOnlyList<(ReaderErrorType errorType, BinaryLogRecordKind recordKind, int count)>
```
KirillOsenkov, line 303: Here I'd actually make a class instead of a ValueTuple

src/StructuredLogger/BinaryLogger/ForwardCompatibilityReadSettings.cs
```diff
@@ -0,0 +1,82 @@
+﻿using System;
+
+namespace Microsoft.Build.Logging.StructuredLogger
+{
+    public delegate bool AllowForwardCompatibilityDelegate(string message);
+
+    public interface IForwardCompatibilityReadSettings
```
KirillOsenkov, line 7: There's no need for this interface :)

src/StructuredLogger/Serialization/Serialization.cs
```diff
@@ -46,9 +47,9 @@ public static class Serialization
         public static Build ReadBuildLog(Stream stream, byte[] projectImportsArchive = null) => BuildLogReader.Read(stream, projectImportsArchive);
         public static Build ReadBinLog(Stream stream, byte[] projectImportsArchive = null) => BinaryLog.ReadBuild(stream, projectImportsArchive);
 
-        public static Build Read(string filePath) => Read(filePath, progress: null);
+        public static Build Read(string filePath) => Read(filePath, progress: null, forwardCompatibilitySettings: null);
```
KirillOsenkov, line 50: I'd pass a simple class ReaderSettings, and on it an enum `UnknownRecordBehavior` `None`, `Error`, `Warning`, `Message`, `ThrowException`

src/StructuredLogger/BinaryLog.cs
```diff
@@ -47,12 +50,25 @@ public static Build ReadBuild(string filePath, Progress progress)
...
             => ReadBuild(stream, progress: null, projectImportsArchive: projectImportsArchive);
 
-        public static Build ReadBuild(Stream stream, Progress progress, byte[] projectImportsArchive = null)
+        public static Build ReadBuild(Stream stream, Progress progress, byte[] projectImportsArchive = null, IForwardCompatibilityReadSettings forwardCompatibilitySettings = null)
         {
-            var eventSource = new BinLogReader();
-
             Build build = null;
             IEnumerable<string> strings = null;
+            ReaderErrorsRegistry readerErrorsRegistry = null;
```
KirillOsenkov, line 57: This feels over-engineered ;)

src/StructuredLogger/BinaryLog.cs
```diff
@@ -133,7 +149,41 @@ public static Build ReadBuild(Stream stream, Progress progress, byte[] projectIm
...
             build.WaitForBackgroundTasks();
 
+            build.IsCompatibilityMode = eventSource.IsCompatibilityMode;
+            build.RecoverableReadingErrors = readerErrorsRegistry?.GetErrors().ToList() ??
+                new List<(ReaderErrorType errorType, BinaryLogRecordKind recordKind, int count)>();
+
             return build;
         }
+
+        private class ReaderErrorsRegistry
```
KirillOsenkov, line 159: Are you sure this level of sophistication is warranted? I wouldn't bother doing any grouping or anything. Just a flat list of simple messages/warnings/errors seems good enough. Or maybe they're not even necessary. All we really want to know is: did we read the whole thing successfully or was there any information we weren't able to handle?

src/StructuredLogViewer/Controls/BuildControl.xaml.cs
```diff
@@ -2056,6 +2056,33 @@ public void DisplayStats()
             //statsRoot.AddChild(histogramNode);
 
             statsRoot.AddChild(new Property { Name = "BinlogFileFormatVersion", Value = Build.FileFormatVersion.ToString() });
+            if (Build.IsCompatibilityMode)
```
KirillOsenkov, line 2059: This is nice but perhaps too much detail? Would be certainly useful to us, but probably too much detail for our users.

Since we already have it, I'm OK with keeping it, just extract method here.

src/StructuredLogViewer.Core/SettingsService.cs
```diff
@@ -299,57 +299,45 @@ public static void SaveCustomArguments(string projectFilePath, string newArgumen
             }
         }
 
-        private static bool enableTreeViewVirtualization = true;
+        private static T Get<T>(ref T backingField)
```
KirillOsenkov, line 302: Thanks for cleaning up settings, it is nice.

src/StructuredLogViewer.Core/SettingsService.cs
```diff
@@ -442,6 +399,7 @@ private static void EnsureSettingsRead()
         const string UseDarkThemeSetting = "UseDarkTheme=";
         const string WindowPositionSetting = "WindowPosition=";
         const string IgnoreEmbeddedFilesSetting = "IgnoreEmbeddedFiles=";
+        const string UseForwardCompatibilitySettings = "UseForwardCompatibility=";
```
KirillOsenkov, line 402: I'd say no need for a setting, just be forward compatible and unconditionally.

KirillOsenkov:
Overall everything looks good. After all the changes on the MSBuild are finalized, do the same changes here and then we can merge this one first. Then I'll publish the viewer and the NuGet and the CLI tools, and then we can merge the MSBuild side.

One concern is that we should ensure the Avalonia version still works. I don't think it needs the GUI or the stats, but at least opening binlogs from the future should work fine.

Ideally if you switch to emitting synthetic errors/warnings instead of custom data on the build or via events, both WPF and Avalonia will get it for free. Just remember to maybe only emit a single error per record type (otherwise we risk reporting thousands of identical errors).

My preferred experience for the consumers of the MSBuild.StructuredLogger NuGet package is that they'll update to the new version and they don't have to change anything else or consume any other API. If they read a binlog from the future with unknown records, emit BuildErrorEventArgs and let them handle it they way they want. If they display the errors someplace, they will see the errors and know to update the NuGet to latest. If they don't see the errors, they probably don't care either, in this way we can just be silent. They could opt in into throwing when the file format is from the future if they want to be notified.

This was a gargantuan amount of work, thanks a ton for driving this!

Don't worry about the CLI, this stuff is still all experimental. Anything is fine.

src/StructuredLogger/BinaryLogger/Postprocessing/ReaderErrorType.cs
paulomorgado: Since there are more unknowns, I'd rather be specific on what unknown is this.

Unless this is a generic unknown where you don't even know what you don't know.

It's always a good practice to have the `0` be something that is invalid, to avoid issues with uninitialized values.

src/StructuredLogger/ObjectModel/Build.cs
paulomorgado: There are performance and semantic differences.

A class is a reference type a `ValueTuple` is a value type. It's the difference between one "big" heap allocation for the backing array and one for each element or just a "big" heap allocation for everything.

C# tuples are meant to be bags of related values that do not represent an entity. They are usually deconstructed on "read".

JanKrivanek:
> My initial gut reaction is that forward compat should be the default for all readers, and if there are compat issues we can emit a synthetic `BuildErrorEventArgs` or `BuildWarningEventArgs`. Maybe consumers could tweak whether they want to see an error, a warning, a message or nothing at all for compat issues.
> 
> So I'd remove things like ForwardCompatibilityExplicitlyConfigured, -fwd, and related things and keep it as simple as possible. Since the current behavior is we throw an exception, literally anything else is an improvement. I'd like users to update to this version of the NuGet and not have to change anything and be automatically opted in.
> 
> The new GUI seems nice, but I'd think it's an overkill (just display a synthetic Error or Warning like I suggest above at the very top of the tree). But since you've already built it, might as well leave it I guess. I think users won't see this often anyway, since the viewer updates when it starts.

I just wanted to flash something e-2-e to have easier discussion over it. No feelings attached to any of it - any of the code can go :-)
I like the suggestion with the synthetic error/warning - lm try that. Making it deafult would make things easier.

Btw. I marked MSBuild changes as not mergable - until new viewer and binlogtools are published. Otherwise it's ready to go - but if you'd have any suggestions or concerns there, definitely still easy to make changes.

src/StructuredLogger/Serialization/Serialization.cs
JanKrivanek: Sounds good.
1 message per whole binlog as well sounds enough.
Just not sure what `UnknownRecordBehavior` would mean?
Shouldn't `Error` be a default; `Warning`, `Message`, `None` be opt-in tunning down of the reporting; and `Throw..` be opt-in of the former behavior? Feels like enough
Or did I misunderstood?

KirillOsenkov:
I left some comments on the MSBuild change - did you see? Nothing critical but some might be beneficial.

src/StructuredLogger/BinaryLogger/Postprocessing/ReaderErrorType.cs
JanKrivanek: @paulomorgado - Kirill was refering to a typo, it'll still mention what unknown it is - basically 'UnkownEventType' -> 'Unk**n**ownEventType'

As for an invalid member - I'd want to avoid it here - as it imposes one extra state to cumbersomely handle on the recepient side, while it's not supposed to happen. Would the current 0 be returned by accident - it'd lead to error anyways.

paulomorgado: Sorry. Missed the whole thing was about a typo.

My comment was that these type of enums should always have a zero value that represents uninitialized/invalid state, because in C# the literal `0` is convertible to any enum value and it's also the default value for the enum.

For most cases (as it seems in this case), that is already the intended value.

JanKrivanek:
@KirillOsenkov - btw. I noticed that replaying binlog in the 'raw' mode is slower and leads to bigger resulting file than with structured replay.
After uncompressing the results are ideticall, but this is still undesired and bit unexpected (contrary to my initiall measurements) - so please hold on with merging - I want to first rootcause that.

src/StructuredLogger/ObjectModel/Build.cs
JanKrivanek: Removing the code alltogether

JanKrivanek:
> @KirillOsenkov - btw. I noticed that replaying binlog in the 'raw' mode is slower and leads to bigger resulting file than with structured replay. After uncompressing the results are ideticall, but this is still undesired and bit unexpected (contrary to my initiall measurements) - so please hold on with merging - I want to first rootcause that.


**tl;dr;:** Fixed :tada: (after long long long evening :-))

1.0.9 (previous released) mean redact time for OrchardCore binlog: 00:00:22.1181747
Current version mean redact time: **00:00:08.7203408**

1.0.9 (previous released) mean log replay time of raw events for OrchardCore binlog: 00:00:22.4146649
Current version mean log replay time of raw events: **00:00:04.9658039**

The issue was hidden in the superfluous flushing of the `BufferedStream` caused by calling `BinaryWriter.BaseStream` (that leads to `Flush`).

KirillOsenkov (approved):
Thank you so much!

KirillOsenkov:
I published https://www.nuget.org/packages/binlogtool/1.0.10 with the latest main branch that includes this PR

KirillOsenkov:
Published https://www.nuget.org/packages/MSBuild.StructuredLogger/2.2.155 as well, @jaredpar if you're feeling adventurous maybe update the package to 2.2.155 and then when MSBuild increments the file format version to 18 hopefully it'll just work and you won't have a build break.

KirillOsenkov:
Published viewer update https://github.com/KirillOsenkov/MSBuildStructuredLog/releases/tag/v2.2.155

PR #745, opened 2024-01-17
[Proposal] Add dumprecords command

### LINKED ISSUE #741: Add binlogtool dumprecords
Add `binlogtool dumprecords` command that writes a .txt file with record kinds and lengths, would be nice for debugging

I have a throw-away prototype here:
https://github.com/KirillOsenkov/MSBuildStructuredLog/tree/dev/kirillo/debuggingLogic

would be nice to see the record kind, length, and maybe (optionally) all the reads performed while reading that record.

Also if the binlog was in the new format >= 18, wondering how much better our diagnostic experience would have been.

@JanKrivanek

JanKrivanek:
Fixes #741

### Context
#741 brought a nice idea. I needed something similar for my investigation - just counts, not sizes - so I stitched something for both.

### Samples

Just summary on OrchardCore (runs for about 0.7s on my machine):

```
>binlogtool dumprecords OrchardCore.binlog --include-total --include-rollup --exclude-details

Total records: 1069320

RecordType, Count, Size[bytes]
==============================
Message, 568034, 38.46
NameValueList, 38359, 60.16
BuildStarted, 1, 12.00
ProjectEvaluationStarted, 417, 35.90
ProjectImported, 64414, 55.11
AssemblyLoad, 22, 55.59
PropertyReassignment, 23492, 41.90
ProjectEvaluationFinished, 417, 1912.71
ProjectStarted, 14300, 56.33
TargetStarted, 49962, 38.80
TaskStarted, 23227, 36.90
TaskFinished, 23227, 34.15
TargetFinished, 49962, 52.04
TaskParameter, 145155, 98.90
EnvironmentVariableRead, 792, 37.90
TargetSkipped, 53035, 58.53
ProjectFinished, 14300, 31.30
TaskCommandLine, 198, 40.51
Error, 5, 46.00
BuildFinished, 1, 14.00
```

Details for binlog from new console:

```
>binlogtool dumprecords

RecordType, Size[bytes]
=======================
Message, 50
Message, 50
NameValueList, 9
BuildStarted, 12
Message, 46
Message, 46
Message, 46
Message, 46
Message, 46
ProjectEvaluationStarted, 34
AssemblyLoad, 54
AssemblyLoad, 54
AssemblyLoad, 54
AssemblyLoad, 54
Message, 41
ProjectImported, 45
ProjectImported, 49
Message, 39
...
```

### Limitations

* Summaries won't run on broken binlogs (just the details list) - feels as acceptable limitation (broken binlogs should be less frequent and just a list of records untill the last complete should then suffice)
* The reported sizes for the new format do not include the record type and the offset itself - so there are 2 bytes per each record not reported (but I'd say that's how we want it)

### Breaking changes

* `BinLogReader.WrapperStream` was removed. It was public, but hopefully nobody was using it. If yes - they can easily switch to `TransparentReadStream`.

### Perf

Summary for OrchardCore runs ~0.178 seconds with the new binlogs (offsets included), ~1.967 seconds with old binlogs. Both after warmup runs.
On full details dumping the difference is not so visible, as majority of the time is spend flushing the records (it's ~56 vs ~58 seconds)

### DIFF
--- a/src/BinlogTool/DumpRecords.cs
+++ b/src/BinlogTool/DumpRecords.cs
@@ -55,7 +55,7 @@ namespace BinlogTool
 
             IEnumerable<(BinaryLogRecordKind, long)> records =
                 new BinLogReader().ChunkBinlog(input)
-                    .Select(r => (r.RecordKind, r.Length));
+                    .Select(r => (r.Kind, r.Length));
 
             int totalRecords = 0;
 
--- a/src/BinlogTool/DumpRecords.cs
+++ b/src/BinlogTool/DumpRecords.cs
@@ -1,0 +1,100 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Build.Logging.StructuredLogger;
+using Microsoft.Build.Logging;
+using StructuredLogger.Utils;
+
+namespace BinlogTool
+{
+    internal static class DumpRecords
+    {
+        public static void Run(List<string> inputs, bool includeTotal, bool includeRollup, bool includeDetails)
+        {
+            if (!inputs.Any())
+            {
+                // Default - current directory
+                inputs.Add(string.Empty);
+            }
+
+            var inputBinlogs = inputs.SelectMany(input => Searcher.FindBinlogs(input, true)).ToList();
+
+            if (!inputBinlogs.Any())
+            {
+                Log.WriteError("No binlogs found.");
+                return;
+            }
+
+            if (inputBinlogs.Count > 1)
+            {
+                Log.WriteLine(
+                    $"Found {inputBinlogs.Count} binlog files. Will provide overview for all. (found files: {(string.Join(',', inputBinlogs))})");
+            }
+
+            foreach (var inputBinlog in inputBinlogs)
+            {
+                if (inputBinlogs.Count > 1)
+                {
+                    Log.WriteLine();
+                    Log.WriteLine($"Overview for {inputBinlog}");
+                    Log.WriteLine();
+                }
+
+                RunSingle(inputBinlog, includeTotal, includeRollup, includeDetails);
+            }
+        }
+
+        public static void RunSingle(string input, bool includeTotal, bool includeRollup, bool includeDetails)
+        {
+            bool needsGreedyEnumerate = includeRollup && includeDetails;
+
+            IEnumerable<(BinaryLogRecordKind, long)> records =
+                new BinLogReader().ChunkBinlog(input)
+                    .Select(r => (r.RecordKind, r.Length));
+
+            int totalRecords = 0;
+
+            if (needsGreedyEnumerate)
+            {
+                records = records.ToList();
+                totalRecords = records.Count();
+            }
+
+            if (includeRollup)
+            {
+                totalRecords = 0;
+                Log.WriteLine("RecordType, Count, Size[bytes]");
+                Log.WriteLine("==============================");
+                foreach (var group in records.GroupBy(r => r.Item1))
+                {
+                    Log.WriteLine($"{group.Key}, {group.Count()}, {group.Sum(r => r.Item2) / (double)group.Count():0.00}");
+                    totalRecords += group.Count();
+                }
+                Log.WriteLine();
+            }
+
+            if (includeDetails)
+            {
+                totalRecords = 0;
+                Log.WriteLine("RecordType, Size[bytes]");
+                Log.WriteLine("=======================");
+                foreach (var record in records)
+                {
+                    Log.WriteLine($"{record.Item1}, {record.Item2}");
+                    totalRecords++;
+                }
+            }
+
+            if (includeTotal)
+            {
+                Log.WriteLine($"Total records: {totalRecords}");
+                Log.WriteLine();
+            }
+        }
+    }
+}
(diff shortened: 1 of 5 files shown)


src/StructuredLogger/BinaryLogger/BinaryLogReplayEventSource.cs
```diff
@@ -57,7 +57,7 @@ internal interface IBinaryLogReplaySource :
     /// by implementing IEventSource and raising corresponding events.
     /// </summary>
     /// <remarks>The class is public so that we can call it from MSBuild.exe when replaying a log file.</remarks>
-    internal sealed class BinaryLogReplayEventSource : EventArgsDispatcher, IBinaryLogReplaySource
+    public sealed class BinaryLogReplayEventSource : EventArgsDispatcher, IBinaryLogReplaySource
```
KirillOsenkov, line 60: I don't think we can make this public because it will then conflict with the one in MSBuild (so consumers referencing both us and MSBuild will have a conflict of both).

src/StructuredLogger/BinaryLogger/BuildEventArgsReader.cs
```diff
@@ -25,7 +25,7 @@ namespace Microsoft.Build.Logging.StructuredLogger
     /// <summary>
     /// Deserializes and returns BuildEventArgs-derived objects from a BinaryReader
     /// </summary>
-    internal partial class BuildEventArgsReader : IBuildEventArgsReaderNotifications, IDisposable
+    public partial class BuildEventArgsReader : IBuildEventArgsReaderNotifications, IDisposable
```
KirillOsenkov, line 28: Same, can't make this public. But you can add a new public API that wraps this call.

src/BinlogTool/DumpRecords.cs
```diff
@@ -0,0 +1,94 @@
...
+                    Log.WriteLine();
+                }
+
+                RunSingle(inputBinlog, includeTotal, includeRollup, includeDetails);
+            }
+        }
+
+        public static void RunSingle(string input, bool includeTotal, bool includeRollup, bool includeDetails)
+        {
+            var reader = BinaryLogReplayEventSource.OpenBuildEventsReader(input);
```
KirillOsenkov, line 54: If we can't have those types public, perhaps add a new public static API on `BinaryLog` that exposes what you want. It already has ReadRecords and friends.

KirillOsenkov:
Does it support the full list of all records (not the summary)? Useful for diffing. Even just the types of records would be valuable so that you can diff two .txt files and see what is the first record where they start to differ.

Would be nice to support formats before 18 too. We can compute records lengths like ReadRecords does, see my draft in #741

JanKrivanek:
@KirillOsenkov - All comments sounds sound and easily accomodable - I'll work on this in a few days.

As for:

> Does it support the full list of all records (not the summary)? Useful for diffing. Even just the types of records would be valuable > so that you can diff two .txt files and see what is the first record where they start to differ.

The ungrouped list is the default mode. I added the optional groupping as that's what I needed for my other investigation.
Though - do you expect the sizes to differ a bit? Would you want to have an option to dump just the types, without sizes?

KirillOsenkov:
When I have to investigate a corrupt binlog, I usually have the good and the bad binlogs. So I dump a flat list of records to see which record is the first broken one. Even the record number and kind would be valuable, but if we can have the size of each record it'd be better (for completeness). If the size of the corrupted record changes, would be nice to know it.

We could deduce the record size from the start positions of records in the stream (this is how we show the progress bar when reading the binlog). But of course in the new format we get the record sizes for free.

src/StructuredLogger/BinaryLogger/BinLogReader.cs
```diff
@@ -272,6 +272,18 @@ public IEnumerable<Record> ReadRecords(string logFilePath)
             return DisposableEnumerable<Record>.Create(ReadRecords(stream), () => stream.Dispose());
         }
 
+        /// <summary>
+        /// Enumerate over all records in the file. For each record reports
+        /// the start position in the stream, length in bytes and the type of record.
+        /// </summary>
+        /// <remarks>Useful for debugging and analyzing binary logs</remarks>
+        public IEnumerable<(BinaryLogRecordKind RecordKind, long Start, long Length)> ChunkBinlog(string logFilePath)
```
KirillOsenkov, line 280: I think let's declare a struct or a record struct instead of a tuple

src/BinlogTool/DumpRecords.cs
```diff
@@ -0,0 +1,100 @@
...
+using System.Threading.Tasks;
+using Microsoft.Build.Logging.StructuredLogger;
+using Microsoft.Build.Logging;
+using StructuredLogger.Utils;
+
+namespace BinlogTool
+{
+    internal static class DumpRecords
+    {
+        public static void Run(List<string> inputs, bool includeTotal, bool includeRollup, bool includeDetails)
```
KirillOsenkov, line 17: not sure we need to suppose multiple binlogs here, but I don't care either way is fine

src/StructuredLogger/BinaryLogger/BinLogReader.cs
KirillOsenkov: we can leave as is

src/BinlogTool/DumpRecords.cs
KirillOsenkov: we can leave as is

JanKrivanek: I got this for free from redactor (where it's helpful)... but true it might not have usage here.
We can remove later (or better extract so the code is on single place)

KirillOsenkov (approved):

KirillOsenkov:
Published https://www.nuget.org/packages/binlogtool/1.0.13

PR #763, opened 2024-03-06
Add native Arm64 support

reflectronic:
The .NET Framework versions of `StructuredLogViewer` and `TaskRunner` will run natively on Arm64. This is using the new [`supportedArchitectures`](https://learn.microsoft.com/en-us/windows/win32/sbscs/application-manifests#supportedarchitectures) element in the application manifest.

### DIFF
--- a/src/StructuredLogViewer/app.manifest
+++ b/src/StructuredLogViewer/app.manifest
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <assembly manifestVersion="1.0" xmlns="urn:schemas-microsoft-com:asm.v1" xmlns:asmv3="urn:schemas-microsoft-com:asm.v3">
-  <assemblyIdentity version="1.0.0.0" name="MyApplication.app"/>
+  <assemblyIdentity version="1.0.0.0" name="MSBuildStructuredLog.StructuredLogViewer" />
   <trustInfo xmlns="urn:schemas-microsoft-com:asm.v2">
     <security>
       <requestedPrivileges xmlns="urn:schemas-microsoft-com:asm.v3">
--- a/src/StructuredLogViewer/app.manifest
+++ b/src/StructuredLogViewer/app.manifest
@@ -1,0 +1,30 @@
+<?xml version="1.0" encoding="utf-8"?>
+<assembly manifestVersion="1.0" xmlns="urn:schemas-microsoft-com:asm.v1" xmlns:asmv3="urn:schemas-microsoft-com:asm.v3">
+  <assemblyIdentity version="1.0.0.0" name="MyApplication.app"/>
+  <trustInfo xmlns="urn:schemas-microsoft-com:asm.v2">
+    <security>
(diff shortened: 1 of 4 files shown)


reflectronic:
image

src/StructuredLogViewer/app.manifest
```diff
@@ -0,0 +1,30 @@
+<?xml version="1.0" encoding="utf-8"?>
+<assembly manifestVersion="1.0" xmlns="urn:schemas-microsoft-com:asm.v1" xmlns:asmv3="urn:schemas-microsoft-com:asm.v3">
+  <assemblyIdentity version="1.0.0.0" name="MyApplication.app"/>
```
KirillOsenkov, line 3: what should the identity be? Perhaps MSBuildStructuredLogViewer.app?

src/TaskRunner/app.manifest
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8"?>
+<assembly manifestVersion="1.0" xmlns="urn:schemas-microsoft-com:asm.v1" xmlns:asmv3="urn:schemas-microsoft-com:asm.v3">
+  <assemblyIdentity version="1.0.0.0" name="MyApplication.app"/>
```
KirillOsenkov, line 3: Maybe TaskRunner.app?

src/StructuredLogViewer/app.manifest
KirillOsenkov: if this is intended to be the assembly name, then perhaps StructuredLogViewer.app?

KirillOsenkov:
Very cool! 

Does the app.manifest file need to be shipped with the app or does it get embedded directly inside the executable?

Let's clarify the assemblyIdentity and then we're good to go!

src/StructuredLogViewer/app.manifest
reflectronic: The article says:
> Uniquely names the application or assembly. Use the following format for the name: `Organization.Division.Name`. For example `Microsoft.Windows.mysampleApp`. Required.

Maybe we can do something like `MSBuildStructuredLog.StrucutredLogViewer`/`MSBuildStructuredLog.TaskRunner`? These names aren't used for anything, we don't use the unmanaged assembly binding, it's just for tidiness

reflectronic:
The app manifest appears to be embedded into the exe, so, no complications there hopefully

src/StructuredLogViewer/app.manifest
KirillOsenkov, line 3:
```suggestion
  <assemblyIdentity version="1.0.0.0" name="MSBuildStructuredLog.StructuredLogViewer" />
```

src/TaskRunner/app.manifest
KirillOsenkov, line 3:
```suggestion
  <assemblyIdentity version="1.0.0.0" name="MSBuildStructuredLog.TaskRunner" />
```

KirillOsenkov (approved):

KirillOsenkov:
Thank you!

PR #767, opened 2024-03-26
Group Context Menu

yuehuang010:
## Consolidate the Context Menu into groups.
The context menu doubled in length.  This change will group common menus into themes, shortening the context menu without removing functionality.

<details><summary>Before</summary>
image
</details> 
<details><summary>After</summary>
image
</details> 

## Small optimization in ShortenText
 - Fix an adding the TrimPrompt when not trimmed.

### DIFF
--- a/src/StructuredLogViewer/Controls/BuildControl.xaml.cs
+++ b/src/StructuredLogViewer/Controls/BuildControl.xaml.cs
@@ -35,6 +35,9 @@ namespace StructuredLogViewer.Controls
         private PreprocessedFileManager preprocessedFileManager;
         private NavigationHelper navigationHelper;
 
+        private MenuItem searchMenuGroup;
+        private MenuItem copyMenuGroup;
+        private MenuItem gotoMenuGroup;
         private MenuItem copyItem;
         private MenuItem copySubtreeItem;
         private MenuItem viewSubtreeTextItem;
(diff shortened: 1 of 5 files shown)


src/StructuredLogViewer/Controls/BuildControl.xaml.cs
```diff
@@ -190,6 +193,10 @@ public BuildControl(Build build, string logFilePath)
             // Build Log
             var contextMenu = new ContextMenu();
             contextMenu.Opened += ContextMenu_Opened;
+            searchMenuGroup = new() { Header = "Search ..." };
+            copyMenuGroup = new() { Header = "Copy ..." };
+            gotoMenuGroup = new() { Header = "Go to ..." };
```
drewnoakes, line 198: nit: menu groups don't usually end with "...".

yuehuang010: Any suggestion?
Just `Search`
or `Search >`

drewnoakes: Looking at menus in other software, the control adds the arrow for you. And your screenshots show the same. So the text would just be e.g. Search.

yuehuang010: It blends in so well, I didn't even notice.  Nice.

src/StructuredLogger/ObjectModel/TextNode.cs
```diff
@@ -4,7 +4,7 @@ public class TextNode : TreeNode
     {
         public string Text { get; set; }
         public string ShortenedText => TextUtilities.ShortenValue(Text);
-        public bool IsTextShortened => Text != ShortenedText;
+        public bool IsTextShortened => Text.Length != TextUtilities.GetShortenLength(Text);
```
KirillOsenkov, line 7: Text may be null I think?

KirillOsenkov:
```suggestion
        public bool IsTextShortened => Text != null && Text.Length != TextUtilities.GetShortenLength(Text);

```

KirillOsenkov (approved):

PR #768, opened 2024-04-01
When opening a project file, look for the preprocess function in the …

yuehuang010:
Check the Evaluation folder for its ProjectEvaluation node to expose the "Preprocess" button.

### DIFF
--- a/src/StructuredLogViewer/Controls/BuildControl.xaml.cs
+++ b/src/StructuredLogViewer/Controls/BuildControl.xaml.cs
@@ -2302,6 +2302,15 @@ Recent (");
                         {
                             // TODO: https://github.com/KirillOsenkov/MSBuildStructuredLog/issues/392
                             evaluation = node.GetNearestParentOrSelf<ProjectEvaluation>();
+
+                            // Check the evaluation folder.
+                            if (evaluation == null && node is Project project)
+                            {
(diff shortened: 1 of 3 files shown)


KirillOsenkov:
I don't quite understand this fix, can you explain what is happening in this scenario? If this was a bug, how do I reproduce?

Great job on fixing the flaky test. It's been bugging me for a while now.

src/StructuredLogViewer/Controls/BuildControl.xaml.cs
```diff
@@ -2302,6 +2302,15 @@ private bool Invoke(BaseNode treeNode)
                         {
                             // TODO: https://github.com/KirillOsenkov/MSBuildStructuredLog/issues/392
                             evaluation = node.GetNearestParentOrSelf<ProjectEvaluation>();
+
```
KirillOsenkov, line 2305: we're in BuildControl, it has a Build, and Build has `FindEvaluation(id)`, perhaps try using that? also instead of doing `node is Project` try perhaps `node.GetNearestParentOrSelf<Project>()`

yuehuang010: This addresses the missing "Preprocess" button when viewing a project (not in evaluation folder).
image

Issue #389, opened 2020-08-31
Support for understanding false conditions, line up values with names


KirillOsenkov:
```
Task "GetReferenceAssemblyPaths" skipped, due to false condition; ('$(TargetFrameworkMoniker)' != '' and ('$(_TargetFrameworkDirectories)' == '' or '$(_FullFrameworkReferenceAssemblyPaths)' == '')) was evaluated as ('.NETFramework,Version=v4.7.2' != '' and ('C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\Common7\IDE\ReferenceAssemblies\Microsoft\Framework\\' == '' or 'C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\Common7\IDE\ReferenceAssemblies\Microsoft\Framework\\' == '')).
```

KirillOsenkov:
Thanks @yuehuang010 for implementing this

PR #769, opened 2024-04-02
Add a message when MSBuild task or target skip due to condition evalu…

yuehuang010:
Add a message explaining why the Condition evaluated to false.  This parses evaluate the condition from the build message and prints which statement in the condition is false.

image


KirillOsenkov:
I've been wanting to fix this in https://github.com/KirillOsenkov/MSBuildStructuredLog/issues/389

KirillOsenkov:
looks very cool!

yuehuang010:
Only had to wait four years.

src/StructuredLogger/Strings/Strings.cs
```diff
@@ -344,14 +344,21 @@ private static string GetPropertyReassignmentText()
             return "^" + text + "$";
         }
 
-        public static Regex CreateRegex(string text, int replacePlaceholders = 0, RegexOptions options = RegexOptions.Compiled)
+        public static Regex CreateRegex(string text, int replacePlaceholders = 0, RegexOptions options = RegexOptions.Compiled | RegexOptions.Singleline, bool capture = false)
         {
-            text = Regex.Escape(text);
+            text = "^" + Regex.Escape(text) + "$";
```
KirillOsenkov, line 349: I think we should check if these two symbols are already present. Also not sure if it's always safe to add these? I hope so?

src/StructuredLogger/Strings/Strings.cs
```diff
@@ -344,14 +344,21 @@ private static string GetPropertyReassignmentText()
...
                 for (int i = 0; i < replacePlaceholders; i++)
                 {
-                    text = text.Replace(@$"\{{{i}}}", ".*?");
+                    if (capture)
+                    {
+                        text = text.Replace(@$"\{{{i}}}", $"(?<group{i}>.*?)");
+                    }
+                    else
+                    {
+                        text = text.Replace(@$"\{{{i}}}", $"(.*?)");
```
KirillOsenkov, line 360: is it safe to add these parentheses `(` and `)`?

src/StructuredLogViewer/Controls/TextViewerControl.xaml.cs
```diff
@@ -159,6 +160,63 @@ public void SetText(string text)
                 return;
             }
 
+            Match matches;
```
KirillOsenkov, line 163: Should this logic perhaps go into BuildControl.xaml.cs -> DisplayText() on line 2371?

Also maybe extract method so that this logic is a separate method.

KirillOsenkov: On the other hand perhaps keep it here, maybe in the future we can do smart highlighting to highlight substrings using the textEditor

KirillOsenkov: still, I'd extract this logic to a separate method

src/StructuredLogger/Strings/Strings.cs
yuehuang010: Added check.  Though I think the Escape would invalidate it.

src/StructuredLogger/Strings/Strings.cs
yuehuang010: Reverted.

src/StructuredLogViewer/Controls/TextViewerControl.xaml.cs
yuehuang010: Moved to a separate function.  I wanted to add Outlining to help visualize the position, but it might be too visually busy.  Maybe in the future, it could have toggle button.

src/StructuredLogger/Strings/Strings.cs
```diff
@@ -344,14 +344,29 @@ private static string GetPropertyReassignmentText()
             return "^" + text + "$";
         }
 
-        public static Regex CreateRegex(string text, int replacePlaceholders = 0, RegexOptions options = RegexOptions.Compiled)
+        public static Regex CreateRegex(string text, int replacePlaceholders = 0, RegexOptions options = RegexOptions.Compiled | RegexOptions.Singleline, bool capture = false)
         {
-            text = Regex.Escape(text);
+            if (text.StartsWith("^") || !text.EndsWith("$"))
```
KirillOsenkov, line 349: I don't quite understand this condition. Can you double-check?

I'd write this as:
```csharp
text = Regex.Escape(text);
if (!text.StartsWith("^") && !text.EndsWith("$"))
{
    text = $"^{text}$";
}
```

ltrzesniewski: I haven't followed the conversation here, but I thought I'd point out that `text.StartsWith("^")` will always be false if `text` is an escaped regex.

KirillOsenkov (approved):

PR #645, opened 2022-12-20
[WIP] Upgrade to Avalonia 11.0 + add Browser project

maxkatz6:
Opening this PR as I thought it should be pretty easy to migrate, but also opens some nice possibilities.
I will push changes to this branch once in a while, updating it with newer Avalonia builds.

What I want to do in this PR
- [x] Upgrade to Avalonia 11.0
- [x] Browser support
- [x] Replace legacy file pickers with sandbox-friendly API (also required updating shared code to use Stream instead of file paths where possible)
- [ ] "Recently opened" links should use bookmarks API instead of saving file paths (for browser support and possibly sandboxed macos or mobile)
- [ ] Replace TreeView with virtualized [TreeDataGrid](https://github.com/AvaloniaUI/Avalonia.Controls.TreeDataGrid) and see how better it is for this application
- [ ] Update Welcome Screen UI to hide features not available on the browser (opening the Project)
- [ ] NET 7 AOT should work with zero changes and configuration, but yet to be tested with this application.
- [ ] Do some codebase research and see if any missing features can be ported from WPF version (like, something that needs TextBlock.Inlines support)
- [ ] Dark mode

Out of scope, but could be done
- Upgrade to Fluent theme (separated PR?)
- Mobile support is possible now, but I don't think this apps needs to have a mobile version (am I wrong?), also needs interface rewritting to be more adaptive

PR #782, opened 2024-05-26
Update Avalonia to latest version, enable basic PublishAot scenario

filipnavara:
Unlike #645 this takes a very minimalistic approach to the Avalonia upgrade. It updates all code to compile and run but it doesn't add Browser support or any optimizations.

For the AOT part, .NET 8 TFM build is added to `StructuredLogger` with couple of annotations. Additionally, the TinyJSON parser is replaced with source generated System.Text.Json deserializer. I didn't update this for the `netstandard2.0` build since I don't want to add dependency on out-of-the-box package.


src/StructuredLogViewer.Avalonia/Controls/SearchAndResultsControl.xaml
```diff
@@ -33,7 +33,7 @@
         <ContentPresenter Name="watermark"
                    Margin="40"
                    HorizontalAlignment="Center"
-                   Content="{Binding WatermarkText}"
+                   Content="WatermarkText"
```
filipnavara, line 36: I am not sure what is this supposed to bind to. It's necessary to provide a proper context for the compiled bindings.

filipnavara:
For reference on macOS ARM64 this produces 30Mb self-contained executable and about 20Mb of native libraries:
```
-rw-r--r--  1 filipnavara  staff       758 Jan 31 13:48 Info.plist
-rwxr-xr-x  1 filipnavara  staff  31278800 May 26 23:51 StructuredLogViewer.Avalonia
drwxr-xr-x  3 filipnavara  staff        96 May 26 23:52 StructuredLogViewer.Avalonia.dsym
-rw-r--r--  1 filipnavara  staff     30924 May 26 23:45 StructuredLogViewer.Core.pdb
-rw-r--r--  1 filipnavara  staff      1702 Jan 31 13:48 StructuredLogViewer.icns
-rw-r--r--  1 filipnavara  staff    174512 May 26 23:51 StructuredLogger.pdb
-rwxr--r--  1 filipnavara  staff   1341120 Apr 25 09:03 libAvaloniaNative.dylib
-rwxr--r--  1 filipnavara  staff   2683424 Apr  9 18:58 libHarfBuzzSharp.dylib
-rwxr--r--  1 filipnavara  staff  15084640 Apr  9 18:58 libSkiaSharp.dylib
```

src/StructuredLogger/Reflector.cs
```diff
@@ -2,11 +2,17 @@
...
+using System.Diagnostics.CodeAnalysis;
+#endif
 
 namespace Microsoft.Build.Logging.StructuredLogger
 {
     public class Reflector
     {
+#if NET8_0_OR_GREATER
+        [DynamicDependency(DynamicallyAccessedMemberTypes.All, typeof(LazyFormattedBuildEventArgs))]
+#endif
```
filipnavara, line 15: This should ideally be rewritten with `UnsafeAccessor` usage.

Imran-imtiaz48 (commented):
Review and Improvements:
Enhanced Initialization:

The new code adds logic to check for the presence of properties or fields to determine which ones to use (useRawArguments and useArgumentsOrFormattedMessage).
Use of UnsafeAccessor:

Introduces the use of UnsafeAccessor for accessing private fields and methods. This provides a clearer and more concise way to handle private members.
Simplification of Methods:

Simplifies the GetArguments method by directly accessing the appropriate fields based on the conditions set in the static constructor.
Conditional Compilation:

Maintains the conditional compilation for different .NET versions (#if NET8_0_OR_GREATER), ensuring compatibility with both older and newer versions.
Extensive Usage of Expression Trees:

Retains the use of expression trees for dynamic method and field accessors in the non-NET8_0_OR_GREATER code block, ensuring backward compatibility.
Suggested Code with Improvements:
csharp
Copy code
using Microsoft.Build.Framework;
#if NET8_0_OR_GREATER
using System.Diagnostics.CodeAnalysis;
using System.Runtime.CompilerServices;
#endif

{
    public class Reflector
    {
#if NET8_0_OR_GREATER
        [DynamicDependency(DynamicallyAccessedMemberTypes.PublicFields | DynamicallyAccessedMemberTypes.PublicMethods, typeof(LazyFormattedBuildEventArgs))]
        static Reflector()
        {
            if (typeof(LazyFormattedBuildEventArgs).GetProperty("RawArguments", BindingFlags.Instance | BindingFlags.NonPublic) is not null)
            {
                useRawArguments = true;
            }
            else if (typeof(LazyFormattedBuildEventArgs).GetField("argumentsOrFormattedMessage", BindingFlags.Instance | BindingFlags.NonPublic) is not null)
            {
                useArgumentsOrFormattedMessage = true;
            }
        }

        static bool useRawArguments;
        static bool useArgumentsOrFormattedMessage;

        [UnsafeAccessor(UnsafeAccessorKind.Method, Name = "get_RawArguments")]
        private extern static object[] GetRawArguments(LazyFormattedBuildEventArgs args);
        [UnsafeAccessor(UnsafeAccessorKind.Field, Name = "argumentsOrFormattedMessage")]
        private extern static ref object GetSetRawArgumentsOrFormattedMessage(LazyFormattedBuildEventArgs args);
        [UnsafeAccessor(UnsafeAccessorKind.Field, Name = "arguments")]
        private extern static ref object[] GetSetRawArguments(LazyFormattedBuildEventArgs args);

        public static object[] GetArguments(LazyFormattedBuildEventArgs args)
        {
            if (useRawArguments)
            {
                return GetRawArguments(args);
            }
            else if (useArgumentsOrFormattedMessage)
            {
                return (object[])GetSetRawArgumentsOrFormattedMessage(args);
            }

            return GetSetRawArguments(args);
        }

        [UnsafeAccessor(UnsafeAccessorKind.Field, Name = "message")]
        private extern static ref string GetSetMessage(BuildEventArgs args);
        public static string GetMessage(BuildEventArgs args)
        {
            return GetSetMessage(args);
        }

        [UnsafeAccessor(UnsafeAccessorKind.Field, Name = "senderName")]
        private extern static ref string GetSetSenderName(BuildEventArgs args);
        public static void SetSenderName(BuildEventArgs args, string senderName)
        {
            GetSetSenderName(args) = senderName;
        }

        [UnsafeAccessor(UnsafeAccessorKind.Field, Name = "timestamp")]
        private extern static ref DateTime GetSetTimestamp(BuildEventArgs args);
        public static void SetTimestamp(BuildEventArgs args, DateTime timestamp)
        {
            GetSetTimestamp(args) = timestamp;
        }
#else
        static Reflector()
        {
            string fieldName = "argumentsOrFormattedMessage";
        @@ -77,18 +134,7 @@ public static void SetTimestamp(BuildEventArgs args, DateTime timestamp)
            timeStampSetter(args, timestamp);
        }

        private static Func<T, R> GetFieldAccessor<T, R>(string fieldName)
        {
            ParameterExpression param = Expression.Parameter(typeof(T), "instance");
            MemberExpression member = Expression.Field(param, fieldName);
        @@ -97,7 +143,7 @@ public static MethodInfo GetEnumerateItemsPerTypeMethod(Type itemDictionary)
            return compiled;
        }

        private static Action<T, R> GetFieldSetter<T, R>(string fieldName)
        {
            ParameterExpression instance = Expression.Parameter(typeof(T), "instance");
            ParameterExpression value = Expression.Parameter(typeof(R), "value");
        @@ -107,5 +153,17 @@ public static MethodInfo GetEnumerateItemsPerTypeMethod(Type itemDictionary)
            Action<T, R> compiled = (Action<T, R>)lambda.Compile();
            return compiled;
        }
#endif

        private static MethodInfo enumerateItemsPerType;
        public static MethodInfo GetEnumerateItemsPerTypeMethod(Type itemDictionary)
        {
            if (enumerateItemsPerType == null)
            {
                enumerateItemsPerType = itemDictionary.GetMethod("EnumerateItemsPerType", BindingFlags.Instance | BindingFlags.NonPublic);
            }

            return enumerateItemsPerType;
        }
    }
}
Summary
The new code leverages advanced features such as UnsafeAccessor and dynamic dependency management, enhancing the flexibility and readability of the Reflector class. It also maintains backward compatibility through conditional compilation, ensuring that the class functions correctly across different .NET versions.

slang25:
@filipnavara this looks great. Now that Avalonia 11.1 is out, are you intending to come back to this?

filipnavara:
Sure, I can update it to 11.1x

filipnavara:
There are still some UI glitches, eg. missing graphics for arrows in tree view. This seems to happen both in the AOT and non-AOT build. It's possible that it happened after the rebase and Avalonia version bump since I don't remember seeing that earlier.

slang25:
Interesting, the arrows seem to be working here after I pulled it down and tried it (both AOT and non-AOT).
 
image

filipnavara:
I pushed a commit that fixed the arrows. 😉

slang25:
I've been using this locally and it all seems good from my perspective 🙂

slang25:
@KirillOsenkov would you be able to review when you get a chance, we can then follow up with some macOS packaging improvements following [this guide](https://avaloniaui.net/blog/the-definitive-guide-to-building-and-deploying-avalonia-applications-for-macos)

KirillOsenkov:
Yes, I'll try to find time within the next couple of weeks, sorry I'm super busy as always.

Do you think theoretically this PR is ready to go or does it need more work?

Should we perhaps look at https://github.com/KirillOsenkov/MSBuildStructuredLog/pull/645 as well and see what can be salvaged from it? I don't know what the state of these two PRs is.

filipnavara:
> Do you think theoretically this PR is ready to go or does it need more work?

It is ready to go. The non-browser bits from #645 are salvaged here.

KirillOsenkov (approved):
Thanks! I really appreciate the help!

KirillOsenkov:
Since this is now merged, could someone with a Mac try out the latest main, try out the instructions to build the Mac version, and see if it all works fine? 

Also our AppVeyor build is producing two zip files with Mac versions for x64 and arm64, do these need updated? Should we start publishing the AOT versions? Sorry I'm not a Mac expert and not sure what's best here.

Every time I publish a release I've been also publishing these two Mac zips in each release.

filipnavara:
> Since this is now merged, could someone with a Mac try out the latest main, try out the instructions to build the Mac version, and see if it all works fine?

It should. I've been primarily testing it on macOS.

> Also our AppVeyor build is producing two zip files with Mac versions for x64 and arm64, do these need updated? Should we start publishing the AOT versions? Sorry I'm not a Mac expert and not sure what's best here.

Ideally we should publish a single universal app. That's a bit tricky with Avalonia. Under normal circumstances you just target `net8.0-macos` TFM and specify `RuntimeIdentifiers=osx-arm64,osx-x64` and the MSBuild magic in Microsoft.macOS SDK does the job for you.

However, historically Avalonia doesn't depend on the macOS API bindings from `net8.0-macos` TFM and uses the universal `net8.0` TFM only. That means you either have to use [3rd-party NuGets](https://github.com/egramtel/dotnet-bundle) to do the bundling job, or follow the [official tutorial](https://docs.avaloniaui.net/docs/deployment/macOS).

Either way, to do proper macOS publishing one needs an Apple Developer account and appropriate certificates. Otherwise you cannot sign the app, and distributing unsigned / ad-hoc signed (ie. signed but without dev certificate) is quite pointless. Such apps cannot be run on modern macOS without resigning or system-wide integrity turned off. Neither of those options are practical.

I do have templates to setup the Apple distribution with a valid developer account, but it still means going through the legalese of creating the account and paying 99 USD per year which may be hard pill to swallow. The alternative is offering source-only builds, or global .NET tool (which is ad-hoc signed on the target machine on installation).

filipnavara:
Small correction to the post above. There are apparently still free Apple Developer accounts. I was not sure if that was still the case, but it's still advertised on [their web](https://developer.apple.com/support/account/).

slang25:
I can confirm the latest AppVeyor zip worked for me with the instructions in the README

slang25:
The free apple developer account thing seems to be a sort of trial, rather than a free offering for open source projects.

I'd be happy to cover the 99 USD per year, through GH sponsorship of whatever is most appropriate.

KirillOsenkov:
ok let me do some research on the apple dev account...

PR #790, opened 2024-07-09
Suggesting target skipped rewording

JanKrivanek:
### Context

Saw a case of a customer confused by target seemingly not being properly ordered. A simplified repro:

image

```xml
<Project DefaultTargets="Target2">
    <Target Name="Target3" BeforeTargets="Target2">
       <Message Text="Target3 executed" Importance="High" />
    </Target>
  
    <Target Name="Target2" Condition="'A' != 'A'">
       <Message Text="Target2 executed" Importance="High" />
    </Target>
</Project>
```

Basically - since target execution is skipped, the ordering is skipped as well (as it doesn't matter in which order is the target skipped). But when looking on a binlog in can feel confusing why the ordering info doesn't seem to match

### Additional notes

I want this as a base for discussion - whether we want to indicate the case in a binlog somehow at all. And if yes - where.
The current proposal might be too broad (possibly irelevant to many cases as there would be no ordering requested).
I do not have a good idea or strong opinion how to surface the info to users - I just feel it would be helpful to surface it

### DIFF
--- a/src/StructuredLogger/BinaryLogger/BuildEventArgsReader.Viewer.cs
+++ b/src/StructuredLogger/BinaryLogger/BuildEventArgsReader.Viewer.cs
@@ -169,7 +169,7 @@ namespace Microsoft.Build.Logging.StructuredLogger
 
                 TargetSkipReason.ConditionWasFalse =>
                     FormatResourceStringIgnoreCodeAndKeyword(
-                        "Target \"{0}\" skipped, due to false condition; ({1}) was evaluated as ({2}).",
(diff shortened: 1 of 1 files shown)


KirillOsenkov:
I'm inclined to won't fix this because we don't want to be deviating from MSBuild here - changing the well-defined string would be confusing (imagine text log uses the original string and binlog shows something else).

Also I wouldn't change anything in the viewer so that we continue to reveal what is really happening at the MSBuild level.

Agreed it's confusing but it's MSBuild that is confusing, and the viewer is just faithful to whatever MSBuild has logged.

JanKrivanek:
My mistake - I didn't know (what a shame :-)) that msbuild uses the identical wording

PR #807, opened 2024-09-04
Add an optimized BinaryReader.

yuehuang010:
WIP


KirillOsenkov:
@JanKrivanek FYI

JanKrivanek:
Complement to https://github.com/dotnet/msbuild/pull/10467 - I still owe another round of review to Felix on that one.

@yuehuang010 - it seems that changes in those PRs should be mirror of each other, so detailed review only on one of those is required - is that correct? Or are there any divergences?

KirillOsenkov:
With Stephen's changes reading the stream is no longer the bottleneck.

PR #810, opened 2024-09-08
remove CustomAppDomainManager

SimonCropp:
i dont think it is used?


KirillOsenkov:
I have used it several times to debug tasks that spawn appdomains. It's rarely used but it was a lot of work to set it all up.

Issue #820, opened 2024-09-17
Use updated DotUtils to avoid lots of expensive Stream.ReadByte calls


stephentoub:
The DotUtils library provides some Stream types used here. Lots of calls end up getting made to those Stream's ReadByte methods, but it hasn't overridden them, such that the (expensive) base implementations are used. I submitted https://github.com/dotutils/streamutils/pull/2 to add those overrides. If/when that's merged and a new package is created that has the updated code, this project should be updated to use the newer version.

KirillOsenkov:
Original:
![image](https://github.com/user-attachments/assets/741d6b46-ebf6-4782-8e29-b2f997bde108)

Low-hanging fruit removed:
![image](https://github.com/user-attachments/assets/116fad1e-b84f-4f2b-abae-8a2a3100bdc2)

This PR:
![image](https://github.com/user-attachments/assets/b4fdec15-4b38-46b7-8d5a-fdb75ac18f2d)

PR #821, opened 2024-09-17
Remove some low-hanging fruit allocation

stephentoub:
@KirillOsenkov asked me to look at why the .NET 8 build was slower than the .NET Framework build (answer: https://github.com/KirillOsenkov/MSBuildStructuredLog/pull/819), and why so much time was being taken up in ReadByte (answer: https://github.com/KirillOsenkov/MSBuildStructuredLog/issues/820, https://github.com/dotutils/streamutils/pull/2). But while looking into it, I did an allocation profile and noticed some low-hanging fruit easily avoided, so this addresses those.

- Avoid closures when calling TreeNode.Find{Last}Child. Added an overload that takes state, and made all call sites use `static` to help flag places state was being closed over.
- Avoid closures in BuildEventArgs.Read. The error paths were closing over state used by the success path, resulting in allocation even on success.
- Changed some methods that were returning interfaces to return the strongly-typed collection instead. Consumers that were iterating through them can now automatically avoid going through an enumerator.
- Avoid params allocations for most call sites to FormatResourceStringIgnoreCodeAndKeyword by adding dedicated overloads for 1/2/3 args.
- Avoid an int boxing allocation (on core) when concatening a string with an int.
- Updated TextUtilities to use more modern APIs, e.g. MemoryExtensions.IndexOfAny, IndexOf, ReplaceLineEndings, etc. Some uses were (accidentally?) using Enumerable.Contains.

### DIFF
--- a/src/BinlogTool/ListTools.cs
+++ b/src/BinlogTool/ListTools.cs
@@ -48,7 +48,7 @@ namespace BinlogTool
                 return null;
             }
 
-            var sourceCommitId = environment.FindChild<Property>(p => p.Name == "SOURCECOMMITID");
+            var sourceCommitId = environment.FindChild<Property>(static p => p.Name == "SOURCECOMMITID");
             if (sourceCommitId != null)
             {
                 return sourceCommitId.Value;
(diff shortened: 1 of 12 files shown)


KirillOsenkov (approved):

KirillOsenkov:
Original:
image

This PR:
image

Allocations are down from 86.8 GB -> 84.4 GB! Very nice!

The time probably needs to be measured several times to be accurate.

PR #824, opened 2024-10-04
add new event type BuildCanceled

JanProvaznik:
Pair PR to https://github.com/dotnet/msbuild/pull/10755
Adds support for new event type serialized there.

also fixes class name BuildSubmissionStartedEventArgs to be uniform with other EventArgs classes

### DIFF
--- a/src/StructuredLogger/Construction/Construction.cs
+++ b/src/StructuredLogger/Construction/Construction.cs
@@ -8,6 +8,7 @@ using System.Linq;
 using Microsoft.Build.Collections;
 using Microsoft.Build.Framework;
 using Microsoft.Build.Framework.Profiler;
+using StructuredLogger.BinaryLogger;
 
 namespace Microsoft.Build.Logging.StructuredLogger
 {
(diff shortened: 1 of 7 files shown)


JanKrivanek (approved):
Thank you

src/StructuredLogger/Construction/Construction.cs
```diff
@@ -512,7 +513,17 @@ public void StatusEventRaised(object sender, BuildStatusEventArgs e)
             {
                 lock (syncLock)
                 {
-                    if (e is ProjectEvaluationStartedEventArgs projectEvaluationStarted)
+                    // If the build was canceled we want to show a message in the build log view.
+                    if (e is BuildCanceledEventArgs buildCanceledEventArgs)
```
KirillOsenkov, line 517: since BuildCanceled is the most rare message, let's put it at the very end of the if chain, so we don't have to test for it first. 

I try to sort these type tests by most common message first, so that in the common case we do less type comparisons.

KirillOsenkov (approved):

KirillOsenkov:
Published:
 * https://github.com/KirillOsenkov/MSBuildStructuredLog/releases/tag/v2.2.356
 * https://www.nuget.org/packages/MSBuild.StructuredLogger/2.2.356
 * https://www.nuget.org/packages/binlogtool/1.0.19

KirillOsenkov:
Please test the new versions with the new binlog format. Thanks!

PR #826, opened 2024-10-17
Add a main menu link to search syntax wiki

vikukush:
I've spent years using binlog viewer almost daily without knowing that it has advanced search syntax.
Adding a link to main menu that points to the wiki to surface this better.
Ideally, the program would pop up some sort of a tip when I'm editing the search field to show advanced command description plus has some minimum highlight of that syntax, but this is not something I can implement here.


KirillOsenkov (approved):
Someone was asking for this just yesterday :) Thanks!

KirillOsenkov:
And yes, search IntelliSense is on the backlog, I just don't know when I will have time to do it.

Issue #823, opened 2024-09-27
[Feature suggestion] 'Search secrets'


JanKrivanek:
### Motivation

People are sometimes cautious about whether they can share their binlogs (rightfully)! They can reduct some secrets today (the set of patterns ideally to be expanded as part of this), but they cannot simply check if there is something to be worried about.

I'd want to add something for just searching secrets - but want to first drag it through discussion before spending time on it.

### UX Proposal

Just a very rough idea :-) I want to collect feedback on what to avoid and what to try when trying to play with possible ux implementations.

![image](https://github.com/user-attachments/assets/773895c3-7704-4805-8366-e13a642b450f)

* It is part of the searching tabs
* The things to be searched are part of the leftmost pane - one can choose what should be searched and hits for chosen section are displayed as part of that section
* The things to search and hits are hierarchical and collapsible
* The found hit are clickable and points to the Log, File content or File List (if hit is in the name of the file)

### Possible future iterations

* The find pane has buttons similar to File -> Redact Secrets, so that the chosen results can be scraped
* Categories selection and expansion is stored in settings
* Ability to specify custom literals and custom patterns
* Ability to search based on high entropy (@KirillOsenkov already has some prototype for this)
* Ability to import custom patterns/literals from file and store them in settings (e.g. 'MySecretProduct' would be always part of the secrets search menu for me if I configure it once)

KirillOsenkov:
My only concern is adding yet another tab, otherwise it looks good! I think there should be a menu File -> Secrets or something, and it would then populate the search pane with what you're suggesting above. Or maybe a dialog.

My question is are there scenarios where we don't want everything enabled? 

@nguerrera FYI if you haven't met @JanKrivanek he's on the MSBuild team and also passionate about redacting secrets among many other things.

KirillOsenkov:
or for starters we could keep it simple and add search keywords such as $secret $token $aws and whatnot. It would work well as an `ISearchExtension`, like `$nuget` for example:
https://github.com/KirillOsenkov/MSBuildStructuredLog/blob/10b05586dc5522d2402937160b499de2fc332411/src/StructuredLogger.Utils/NuGetSearch.cs#L48

JanKrivanek:
I like the ideas to simplify this - especially with search term.
[...]

PR #829, opened 2024-10-30
Add secrets search and bump DotUtils.MsBuild.BinlogRedactor.SensitiveDataDetector version

YuliiaKovalova:
## Fixes
https://github.com/KirillOsenkov/MSBuildStructuredLog/issues/823

## Context
This change adds a support of a new search key word "**$secret**" that allows to detect all the suspicious tree entries, based on the capabilities of this library [MSBuild.BinlogRedactor](https://github.com/dotutils/MSBuild.BinlogRedactor).

MSBuild.BinlogRedactor is already used for secrets reduction functionality, but it this context it's possible to check the presence of the secrets on fly
{4CD40D35-CD93-4195-B83B-30873E35C1D2}

**$secret not**([SensitiveDataKind](https://github.com/dotutils/MSBuild.BinlogRedactor/blob/main/src/DotUtils.MsBuild.SensitiveDataDetector/SensitiveDataDetectorFactory.cs)) statement is supported here:
{E743D083-2878-4655-AACF-711D9A8E6610}

It's possible to use the new keyword on 2 tabs:
 **"Search Log"** 
{12CCFBFD-AC63-4A06-B1F7-86948D7B721A}

**"Find in Files"**
{ED9BC1B4-EBC7-4979-B8E2-097B9A01AB53}

## Changes made
 A new SecretsSearch class was implemented based on ISearchExtension interface.
To make this functionality available on "Find in Files" tab, the method was extended https://github.com/KirillOsenkov/MSBuildStructuredLog/blob/6e4fc91055cc354e57684649d10a03e0e826b072/src/StructuredLogViewer/Controls/BuildControl.xaml.cs#L1022. There was an idea to use factory for handing this case, but we agreed with to postpone this and get back on demand.
[msbuild_logWithFalseSecrets.zip](https://github.com/user-attachments/files/17588093/msbuild_logWithFalseSecrets.zip)

### DIFF
--- a/src/StructuredLogViewer/Controls/BuildControl.xaml.cs
+++ b/src/StructuredLogViewer/Controls/BuildControl.xaml.cs
@@ -16,6 +16,7 @@ using System.Windows.Media;
 using System.Windows.Threading;
 using System.Xml;
 using DotUtils.MsBuild.SensitiveDataDetector;
+using Microsoft.Build.Experimental.ProjectCache;
 using Microsoft.Build.Framework;
 using Microsoft.Build.Logging.StructuredLogger;
 using Microsoft.Language.Xml;
@@ -1029,9 +1030,17 @@ Recent (");
                     return null;
                 }
 
-                if (searchText == "$secret")
+                if (!string.IsNullOrEmpty(searchText) && searchText.StartsWith("$secret"))
                 {
-                    List<SecretDescriptor> searchResults = secretsSearch.GetSecrets(file.Value.Text);
+                    var word = searchText.Replace("$secret", string.Empty).Trim();
+                    NodeQueryMatcher notMatcher = null;
+                    if (word.StartsWith("not(", StringComparison.OrdinalIgnoreCase) && word.EndsWith(")"))
+                    {
+                        word = word.Substring(4, word.Length - 5);
+                        notMatcher = new NodeQueryMatcher(word);
+                    }
+
+                    var searchResults = secretsSearch.SearchSecrets(file.Value.Text, notMatcher, maxResults);
                     if (searchResults.Count > 0)
                     {
                         results.Add((file.Key, searchResults.Select(sr => (sr.Line - 1, sr.Secret))));
--- a/src/StructuredLogViewer/Controls/BuildControl.xaml.cs
+++ b/src/StructuredLogViewer/Controls/BuildControl.xaml.cs
@@ -18,7 +18,6 @@ using System.Xml;
 using DotUtils.MsBuild.SensitiveDataDetector;
 using Microsoft.Build.Framework;
 using Microsoft.Build.Logging.StructuredLogger;
-using Microsoft.Build.SensitiveDataDetector;
 using Microsoft.Language.Xml;
 using Mono.Cecil;
 using StructuredLogViewer.Core.ProjectGraph;
--- a/src/StructuredLogViewer/Controls/BuildControl.xaml.cs
+++ b/src/StructuredLogViewer/Controls/BuildControl.xaml.cs
@@ -15,8 +15,10 @@ using System.Windows.Input;
 using System.Windows.Media;
 using System.Windows.Threading;
 using System.Xml;
+using DotUtils.MsBuild.SensitiveDataDetector;
 using Microsoft.Build.Framework;
 using Microsoft.Build.Logging.StructuredLogger;
+using Microsoft.Build.SensitiveDataDetector;
 using Microsoft.Language.Xml;
 using Mono.Cecil;
 using StructuredLogViewer.Core.ProjectGraph;
@@ -79,6 +81,8 @@ namespace StructuredLogViewer.Controls
 
         private PropertiesAndItemsSearch propertiesAndItemsSearch;
 
+        private SecretsSearch secretsSearch;
+
         public BuildControl(Build build, string logFilePath)
         {
             InitializeComponent();
@@ -137,6 +141,7 @@ namespace StructuredLogViewer.Controls
             propertiesAndItemsControl.WatermarkDisplayed += UpdatePropertiesAndItemsWatermark;
             propertiesAndItemsControl.RecentItemsCategory = "PropertiesAndItems";
 
+            secretsSearch = new SecretsSearch(build);
             SetProjectContext(null);
 
             VirtualizingPanel.SetIsVirtualizing(treeView, SettingsService.EnableTreeViewVirtualization);
@@ -662,7 +667,8 @@ Right-clicking a project node may show the 'Preprocess' option if the version of
             "$csc",
             "$rar",
             "$import",
-            "$noimport"
+            "$noimport",
+            "$secret"
         };
 
         private static Inline MakeLink(string query, SearchAndResultsControl searchControl, string before = " \u2022 ", string after = "\r\n")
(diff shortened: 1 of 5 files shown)


src/StructuredLogger.Utils/SecretsSearch.cs
```diff
@@ -0,0 +1,140 @@
...
+            return secrets.Values.SelectMany(v => v).Take(maxResults).ToList();
+        }
+
+        private IEnumerable<SearchResult> ScanForSecrets(IEnumerable<string> stringsPool, IReadOnlyDictionary<SensitiveDataKind, ISensitiveDataDetector> detectors, int maxResults)
+        {
+            var results = new List<SearchResult>();
+
+            foreach (var text in stringsPool)
+            {
+                if (_resultCache.TryGetValue(text, out List<SearchResult> cachedResults))
```
KirillOsenkov, line 78: Do we need the cache?

src/StructuredLogger.Utils/SecretsSearch.cs
```diff
@@ -0,0 +1,140 @@
...
+                    }
+
+                    continue;
+                }
+
+                var secrets = DetectSecrets(text, detectors);
+
+                if (secrets.Any())
+                {
+                    var nodes = _search.FindNodes(text, default);
```
KirillOsenkov, line 93: there are two ways to search: this one and the one using SearchIndex.
I think the SearchIndex one should be faster:
https://github.com/KirillOsenkov/MSBuildStructuredLog/blob/fa3c09e998926b1f9cff90ea87456901ff6858d0/src/StructuredLogViewer/Controls/BuildControl.xaml.cs#L90-L106

src/StructuredLogger.Utils/SecretsSearch.cs
```diff
@@ -0,0 +1,140 @@
...
+                    {
+                        break;
+                    }
+
+                    continue;
+                }
+
+                var secrets = DetectSecrets(text, detectors);
+
+                if (secrets.Any())
```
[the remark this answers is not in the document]
YuliiaKovalova: I like it, done!

src/StructuredLogger.Utils/SecretsSearch.cs
YuliiaKovalova: it works better for sure, done

src/StructuredLogViewer/Controls/BuildControl.xaml.cs
```diff
@@ -1024,11 +1030,30 @@ private object FindInFiles(string searchText, int maxResults, CancellationToken
...
-                var resultsInFile = haystack.Find(searchText);
-                if (resultsInFile.Count > 0)
+                if (!string.IsNullOrEmpty(searchText) && searchText.StartsWith("$secret"))
                 {
-                    results.Add((file.Key, resultsInFile.Select(lineNumber => (lineNumber, haystack.GetLineText(lineNumber)))));
+                    var word = searchText.Replace("$secret", string.Empty).Trim();
+                    NodeQueryMatcher notMatcher = null;
+                    if (word.StartsWith("not(", StringComparison.OrdinalIgnoreCase) && word.EndsWith(")"))
+                    {
+                        word = word.Substring(4, word.Length - 5);
```
JanKrivanek, line 1039: Super nit pick
```suggestion
                        word = word.Substring("not(".Length, word.Length - "not()".Length);

```
or a comment explaining the magic numbers

src/StructuredLogger.Utils/SecretsSearch.cs
```diff
@@ -0,0 +1,130 @@
...
+                    if (kv.Value.Any())
+                    {
+                        results[kv.Key] = kv.Value;
+                    }
+                }
+            }
+
+            if (results.Any())
+            {
+                _secretCache[text] = results;
```
JanKrivanek, line 112: Is this guaranteed to be called single-threaded from the UI?

JanKrivanek (approved):
I love this!

JanKrivanek:
Can this be added to list of sample searches?

https://github.com/KirillOsenkov/MSBuildStructuredLog/blob/fa3c09e998926b1f9cff90ea87456901ff6858d0/src/StructuredLogViewer/Controls/BuildControl.xaml.cs#L633-L646

image

src/StructuredLogger.Utils/SecretsSearch.cs
YuliiaKovalova: no idea, @KirillOsenkov could you advice?

KirillOsenkov: It is called on a background thread, but there's no parallelism involved, so no need for a lock. When in doubt you can put a breakpoint in the code and see from which thread it is called:

image

src/StructuredLogViewer/Controls/BuildControl.xaml.cs
```diff
@@ -137,6 +141,7 @@ public BuildControl(Build build, string logFilePath)
             propertiesAndItemsControl.WatermarkDisplayed += UpdatePropertiesAndItemsWatermark;
             propertiesAndItemsControl.RecentItemsCategory = "PropertiesAndItems";
 
+            secretsSearch = new SecretsSearch(build);
```
KirillOsenkov, line 144: build.Analyze() will add a SecretsSearch to the build itself, perhaps no need to do it here? when you create new SecretsSearch() in the other location, save the instance to the Build:
https://github.com/YuliiaKovalova/MSBuildStructuredLog/blob/6e4fc91055cc354e57684649d10a03e0e826b072/src/StructuredLogViewer/MainWindow.xaml.cs#L695

and then you can access it off the Build.SecretsSearch

KirillOsenkov: basically make sure that the constructor is called in a single place

src/StructuredLogViewer/Controls/BuildControl.xaml.cs
```diff
@@ -1024,11 +1030,30 @@ private object FindInFiles(string searchText, int maxResults, CancellationToken
                     return null;
                 }
 
-                var haystack = file.Value;
-                var resultsInFile = haystack.Find(searchText);
-                if (resultsInFile.Count > 0)
+                if (!string.IsNullOrEmpty(searchText) && searchText.StartsWith("$secret"))
```
KirillOsenkov, line 1033: this condition will run for every file, and it doesn't change. How about we extract a bool above the foreach and just check it here?

KirillOsenkov:
Very nice! I left a couple small comments but otherwise it's good to go!

src/StructuredLogViewer/Controls/BuildControl.xaml.cs
```diff
@@ -1024,11 +1030,30 @@ private object FindInFiles(string searchText, int maxResults, CancellationToken
                     return null;
                 }
 
-                var haystack = file.Value;
-                var resultsInFile = haystack.Find(searchText);
-                if (resultsInFile.Count > 0)
+                if (!string.IsNullOrEmpty(searchText) && searchText.StartsWith("$secret"))
                 {
-                    results.Add((file.Key, resultsInFile.Select(lineNumber => (lineNumber, haystack.GetLineText(lineNumber)))));
+                    var word = searchText.Replace("$secret", string.Empty).Trim();
```
KirillOsenkov, line 1035: instead of manually parsing the query, try just doing `var matcher = new NodeQueryMatcher(searchText)`

the resulting matcher will have the `TypeKeyword`, the `Terms` and the `NotMatchers`.

KirillOsenkov: also keep in mind that you can have multiple not matchers, so you may want to pass multiple to SearchSecrets()

KirillOsenkov: oh, and parsing the searchText can also be done outside the foreach loop, to avoid repetitive work

src/StructuredLogViewer/Controls/BuildControl.xaml.cs
```diff
@@ -1023,24 +1022,19 @@ private object FindInFiles(string searchText, int maxResults, CancellationToken
         {
             var results = new List<(string, IEnumerable<(int, string)>)>();
 
+            NodeQueryMatcher notQueryMatcher = new NodeQueryMatcher(searchText);
```
KirillOsenkov, line 1025: did you mean `nodeQueryMatcher`?

KirillOsenkov: I fixed it in main

JanKrivanek:
Works like a charm!
Plus the new patterns are now available in the `binlogtool redact` as well:

image

JanKrivanek:
@YuliiaKovalova  - do you want to promete the functionality on the viewer help? Probably here?: https://github.com/KirillOsenkov/MSBuildLog/blob/6f87e14be0eb1a7cfe7c40e912b96b9ce0cf0301/index.html#L185-L191

KirillOsenkov:
One todo would be to investigate adding parallelism to secret detection because right now it's really slow on real-life binlogs, even moderately sized. Takes over two minutes on a 5 MB binlog I use often. I think a judicious use of Task.Run() will really make things better here. You can maybe spawn one task per string, and one task per file, then just foreach the tasks and do task.Wait() on each of them. No need for async as only one thread will be blocked (the one where the computation happens)

PR #842, opened 2024-12-15
MacOS signing

maxkatz6:
Steps:
- [x] Minimal sign scripts, figuring out how to use Cake.
- [x] Update AppVeyor to run mac build on MacOS machine. This will automatically solve "chmod +x" problem too.
- [x] Use p12 and password from the secrets/env.
- [ ] Validate that signing actually works
- [ ] Try writing notarizing script (it might timeout on CI with JIT entitlements)

Optional, to be done in other PRs:
- [ ] Use NativeAOT in the release builds, remove JIT entitlements, try notarizing again
- [ ] Create DMG file

Closes #686

PR #845, opened 2024-12-17
Update macOS istallation readme

maxkatz6:
With https://github.com/KirillOsenkov/MSBuildStructuredLog/pull/842 PR successfully signing app bundle, we can remove couple of unnecessary steps. This PR also adds one more step to allow unverified developer ID, until we have proper notarization process set-up.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,15 +22,13 @@ The app updates automatically via [Squirrel](https://github.com/Squirrel/Squirre
 
 ## Installing the Avalonia version on Mac:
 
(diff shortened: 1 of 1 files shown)


akoeplinger:
FYI I tried the .dmg from the build after https://github.com/KirillOsenkov/MSBuildStructuredLog/pull/853 was merged and was able to install without any additional steps, great job!

maxkatz6:
Yes, since we have DMG now, whole workaround instruction can be removed.

Issue #795, opened 2024-08-06
Analyzer / Generator summary doesn't display on non-English builds


jaredpar:
The analyzer and generator summary nodes are created by looking for the [following strings][viewer match] in a `Message` node in the binary log: 

- `Total analyzer execution time:`
- `Total generator execution time:`

That marks the beginning of the summary output and the viewer will parse out the nodes below that. The problem is that those strings in the compiler are [localizable strings][compiler string]. That means on non-English builds they will have different values. The strings in [the  viewer][viewer string] are hard coded. That means they won't match and the viewer won't produce summary nodes for analyzers or generators. 

I'm not sure what the best way is to fix this. The compiler needs to emit localized messages here as it's part of the standard compiler output that is localized. There is no reasonable way for the viewer to match this across locale's. 

Open to suggestions on how to best approach this problem.

[viewer string]: https://github.com/KirillOsenkov/MSBuildStructuredLog/blob/21f0d96143916bcb0472dc347f04059375869f2c/src/StructuredLogger/Strings/Strings.cs#L602
[viewer match]: https://github.com/KirillOsenkov/MSBuildStructuredLog/blob/21f0d96143916bcb0472dc347f04059375869f2c/src/StructuredLogger/Analyzers/CscTaskAnalyzer.cs#L21
[compiler string]: https://github.com/dotnet/roslyn/blob/1d424c126a07054f0b84b6ae28c13c7d4336db62/src/Compilers/Core/Portable/CodeAnalysisResources.resx#L529

KirillOsenkov:
if you have a sample binlog, can you send it to me?

KirillOsenkov:
This should have been fixed with https://github.com/KirillOsenkov/MSBuildStructuredLog/pull/857

PR #857, opened 2025-02-25
Fix analyzer and generator summary on non english system

kvpt:
Fix issue #795.
To do that I updated the Strings resources with the ResourceCreator project (had to set x86 platform explicitly to load MSBuild assembly).
By doing this, I noticed that some resources have changed, not sure what to do about this.
I also added several resources names that were missing.

Here a before/after.
image

### DIFF
--- a/src/ResourcesGenerator/ResourcesGenerator.csproj
+++ b/src/ResourcesGenerator/ResourcesGenerator.csproj
@@ -2,7 +2,8 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>net472</TargetFramework>
+    <TargetFramework>net48</TargetFramework>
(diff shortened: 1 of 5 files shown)


src/ResourcesGenerator/ResourcesGenerator.csproj
```diff
@@ -2,7 +2,8 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>net472</TargetFramework>
+    <TargetFramework>net48</TargetFramework>
```
KirillOsenkov, line 5: let's keep this at net472

KirillOsenkov: I'll fix it myself

KirillOsenkov (approved):
Thanks!

PR #862, opened 2025-03-08
remove hardcoded RIDs

kasperk81:
these are not necessary and make it work on any platform supported by runtime like linux-arm64 without hardcoding them
```
dotnet publish -c Release --self-contained
```

will publish for current RID. with `-r other-rid` will publish for other-rid (cross build)

### DIFF
--- a/build-macos.cake
+++ b/build-macos.cake
@@ -20,7 +20,8 @@ var buildDirs =
 var netCoreProject = new {
         Path = $"{netCoreAppsRoot}/{netCoreApp}",
         Name = netCoreApp,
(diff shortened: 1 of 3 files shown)


KirillOsenkov:
CI is failing

KirillOsenkov:
@maxkatz6 a quick review whenever you have time

KirillOsenkov (approved):

maxkatz6:
It looks fine. 
Specifically on macOS we could just build without any rid specified, and it would create an universal binary for x64 and arm64. But I don't remember if it's AOT only or not.

PR #866, opened 2025-05-09
Added "Sort children by duration" to tree nodes

mehrabiworkmail:
Having this feature makes the tool more useful for investigating build slowdowns.


KirillOsenkov (approved):
Very nice!

PR #872, opened 2025-06-02
Fix binlog argument passing for output file

stan-sz:


src/BinlogTool/Program.cs
```diff
@@ -83,15 +83,11 @@ binlogtool search *.binlog search string
                 return 0;
             }
 
-            if (string.Equals(firstArg, "compilerinvocations", StringComparison.OrdinalIgnoreCase))
+            if (args.Length >= 2 && string.Equals(firstArg, "compilerinvocations", StringComparison.OrdinalIgnoreCase))
             {
-                string binlog = null;
+                string binlog = binlog = args[1];
```
KirillOsenkov, line 88: = binlog = binlog is duplicated

KirillOsenkov: I'll fix it myself

PR #875, opened 2025-06-18
Add Tooltip for MSBuild Properties in Preprocess View.

yuehuang010:
### Add Tooltip for MSBuild Properties in Preprocess View.
 - Add a simple tooltip that quickly shows the evaluated value for the property and the previous evaluations.
 - This feature saves time as it avoids a search in the "Properties and Item" tab.
 - I would like to add item group too, but the visual tends to be too long.  The tooltip popup should first richer.  

image

### DIFF
--- a/src/StructuredLogViewer/Controls/DocumentOutline.cs
+++ b/src/StructuredLogViewer/Controls/DocumentOutline.cs
@@ -1,4 +1,10 @@
-using System;
+﻿using System;
+using System.Linq;
+using System.Text;
+using System.Windows.Controls;
+using System.Windows.Controls.Primitives;
+using System.Windows.Input;
+using ICSharpCode.AvalonEdit;
 using Microsoft.Build.Logging.StructuredLogger;
 using StructuredLogViewer.Controls;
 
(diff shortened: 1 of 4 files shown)


src/StructuredLogger/Strings/Strings.cs
```diff
@@ -327,32 +327,32 @@ private static string GetPropertyReassignmentText()
...
+                        .Replace(@"$({0})=“{1}”(", @"\$\((?<Name>\w+)\)=“(?<NewValue>.*)”\(")
+                        .Replace(@"{2}”)", @"(?<OrgValue>.*)”\)")
                         .Replace("{3}", @"(?<File>.*) \((?<Line>\d+),(?<Column>\d+)\)");
                     break;
                 default:
                     text = text
-                        .Replace(@"$({0})=""{1}"" (", @"\$\((?<Name>\w+)\)="".*"" \(")
-                        .Replace(@"{2}"")", @".*""\)")
+                        .Replace(@"$({0})=""{1}"" (", @"\$\((?<Name>\w+)\)=""(?<NewValue>.*)"" \(")
+                        .Replace(@"{2}"")", @"(?<OrgValue>.*)""\)")
```
KirillOsenkov, line 355: did you mean OldValue instead of OrgValue?

src/StructuredLogViewer/Controls/DocumentOutline.cs
```diff
@@ -22,6 +28,158 @@ public void Install(TextViewerControl textViewerControl)
...
+                {
+                    var propertyEntry = propertyFolder.Children.FirstOrDefault(p => p.Title == title);
+                    if (propertyEntry is Property property)
+                    {
+                        content.Append($"{title} =\n{property.Value.NormalizePropertyValue()}");
+                    }
+                }
+
+                // Search the "Property reassignment" folder.
+                var prFolder = this.PreprocessContext.Evaluation.Children.FirstOrDefault(p => p.Title == "Property reassignment");
```
KirillOsenkov, line 70: You can use the Strings.PropertyReassignment I think

src/StructuredLogger/Strings/Strings.cs
yuehuang010: Org == Original, but `Old` is a better match for the 3 letter `New`.

PR #879, opened 2025-07-03
Fix RecordKind populating

### LINKED ISSUE #877: `BinaryLog.ReadRecords` reports `Kind` as `EndOfFile` for almost all records
Steps to reproduce:

```ps1
dotnet new console
dotnet build -bl # to create sample binlog
dotnet add package MSBuild.StructuredLogger --version 2.2.472
```

Replace `Program.cs` with:

```cs
using Microsoft.Build.Logging.StructuredLogger;

var records = BinaryLog.ReadRecords("msbuild.binlog").ToList();
Console.WriteLine($"Records: {records.Count}");
Console.WriteLine($"Kinds: {string.Join(", ", records.CountBy(r => r.Kind))}");
Console.WriteLine($"Example: {records.First(r => r.Args?.Message == "EvaluationStarted")}");
```

```ps1
dotnet run
```

Observed output:

```txt
Records: 3893
Kinds: [EndOfFile, 3892], [ProjectImportArchive, 1]
```

Expected: some other kinds as well, for example `BinaryLogRecordKind.ProjectEvaluationStarted` for `records.First(r => r.Args?.Message == "EvaluationStarted").Kind`.

JanKrivanek:
Fix #877

## Background

`Record.Kind` was not populated in `ReadRecordsFromDecompressedStream` execution path - so added it there

## Drawback

Missing test(s) on `ReadRecords`

This change will require the `ToBinaryLogRecordKind` to be kept properly in sync with possible new added EventArgs - otherwise `ReadRecords` might throw otherwise. So test would be nice to prevent about forgotting. (btw. this affects api only - not the GUI - it uses different reading methods)

WDYT @KirillOsenkov? Would you preffer this to already added fix (https://github.com/KirillOsenkov/MSBuildStructuredLog/commit/78d683efc99fa0c1b8eac14a5d21dfdbecc8ecfe)?

### DIFF
--- a/src/StructuredLogger/BinaryLogger/BinLogReader.cs
+++ b/src/StructuredLogger/BinaryLogger/BinLogReader.cs
@@ -7,6 +7,7 @@ using System.IO;
 using System.IO.Compression;
 using System.Linq;
 using Microsoft.Build.Framework;
+using StructuredLogger.BinaryLogger;
 
 namespace Microsoft.Build.Logging.StructuredLogger
 {
@@ -319,7 +320,7 @@ namespace Microsoft.Build.Logging.StructuredLogger
             else
             {
                 return ReadRecordsFromDecompressedStream(reader, true)
-                    .Select(r => new RecordInfo(r.Args != null ? ToBinaryLogRecordKind(r.Args) : r.Kind, r.Start, r.Length));
+                    .Select(r => new RecordInfo(r.Kind, r.Start, r.Length));
             }
         }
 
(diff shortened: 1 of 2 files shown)


KirillOsenkov:
This looks good, my only potential concern is the type switch might be more expensive than just passing the integer out.

But maybe my intuition is wrong and the cost will be completely drowned in disk access and decompression. 

Would you mind measuring ReadRecords for a large binlog with whatever is in main vs. your PR? Remember to warm up first, then measure. If there's no noticeable difference I'm happy to merge.

JanKrivanek:
You were right @KirillOsenkov! (well - how could I doubt about your perf instinct ;-D)
There is a small - but measurable - perf hit by this PR method. So I'm abandoning this for now

It was fun experiment though :)

## Data

| LogName | Records # |
|------------|------------|
| msbuild-console.binlog cnt | 5928 |
| msbuild-10-projs.binlog cnt | 55952 |
| msbuild-50-projs.binlog cnt | 502567 |
| msbuild-100-projs.binlog cnt | 395290 |

## Main

| Method          | Mean       | Error     | StdDev     |
|---------------- |-----------:|----------:|-----------:|
| ReadConsoleLog  |   5.925 ms | 0.1182 ms |  0.1213 ms |
| Read10ProjsLog  |  56.885 ms | 1.1312 ms |  2.2329 ms |
| Read50ProjsLog  | 456.204 ms | 9.0176 ms | 18.4205 ms |
| Read100ProjsLog | 238.662 ms | 4.6952 ms |  7.1700 ms |

## This PR

| Method          | Mean       | Error     | StdDev    |
|---------------- |-----------:|----------:|----------:|
| ReadConsoleLog  |   6.221 ms | 0.0824 ms | 0.0771 ms |
| Read10ProjsLog  |  59.692 ms | 1.0683 ms | 1.8711 ms |
| Read50ProjsLog  | 469.211 ms | 4.9124 ms | 4.3547 ms |
| Read100ProjsLog | 252.555 ms | 4.9209 ms | 5.8580 ms |

Issue #885, opened 2025-08-13
No public event args raised for property changes


ericstj:
I was working on a tool that tries to extract properties from binlogs.  I can capture the initial properties from the evaluation, but property changes during target execution are not raised with the property info as a public type.

https://github.com/KirillOsenkov/MSBuildStructuredLog/pull/839/files#diff-5850999507dea96dddb234354cf875f2e8ba46b9f7d0e0dd1019bbf137e3e60bR1376
This change seems to report that information as an `ExtendedPropertyReassignmentEventArgs` which is an internal type.  As a result, I can't capture the property change.

If that type derived from the public type `PropertyReassignmentEventArgs` then I'd still be able to get at the properties I needed.

I can workaround by parsing out the message as a hack, but please bring back public access to property changes.  @YuliiaKovalova 

Here's a basic repro:
```xml
<Project>
  <PropertyGroup>
    <MyProp>initial</MyProp>
  </PropertyGroup>

  <Target Name="ChangeMyProp" BeforeTargets="Build">
    <PropertyGroup>
      <MyProp>changed</MyProp>
    </PropertyGroup>
  </Target>

  <Target Name="Build" />

</Project>
```

`dotnet msbuild msbuildTest.csproj /t:Build /bl`

PR #887, opened 2025-08-14
Make ExtendedPropertyReassignmentEventArgs & ExtendedPropertyInitialValueSetEventArgs public

YuliiaKovalova:
## Fixes 
https://github.com/KirillOsenkov/MSBuildStructuredLog/issues/885

## Changes made

This code makes ExtendedPropertyReassignmentEventArgs & ExtendedPropertyInitialValueSetEventArgs public and they derive from the corresponding MSBuild types.


src/StructuredLogger/BinaryLogger/ExtendedPropertyReassignmentEventArgs.cs
```diff
@@ -28,26 +28,8 @@ public ExtendedPropertyReassignmentEventArgs(
             string helpKeyword = null,
             string senderName = null,
             MessageImportance importance = MessageImportance.Low)
-            : base(subcategory: null, code: null, file: file, lineNumber: line, columnNumber: column, 0, 0, message, helpKeyword, senderName, importance)
+            : base(propertyName, previousValue, newValue, location: message, message)
```
ericstj, line 31: Is there a reason why we can't just call this constructor directly?  Why do we need our own type?

KirillOsenkov (approved):

ericstj (commented):
Thanks!

PR #895, opened 2025-09-12
Add "Direct references only" option to graph control

drewnoakes:
The size of the target graph is often so large that related nodes aren't visible within the viewport.

This change adds a checkbox to the toolbar that causes nodes that aren't directly referenced from the selected node. Clicking on one of those nodes will refocus the graph around that node, adding any newly needed nodes and hiding any no longer directly referenced.

Screencast (includes command from #894, though that's not in this PR):

direct-references-only

### DIFF
--- a/src/StructuredLogViewer/Controls/GraphControl.cs
+++ b/src/StructuredLogViewer/Controls/GraphControl.cs
@@ -116,6 +116,34 @@ public class GraphControl
         }
     }
 
+    private bool directReferencesOnly;
+    private Vertex focusVertex;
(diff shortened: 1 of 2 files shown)


KirillOsenkov:
maybe another checkbox ("show transitive closure only") or something

drewnoakes:
Ok, so one checkbox to support both directions (one option for both cones rather than two options for separate cones). Will give it a crack.

drewnoakes:
https://github.com/KirillOsenkov/MSBuildStructuredLog/pull/895#pullrequestreview-3221831667

PR #900, opened 2025-09-19
Ctrl + Shift + F focuses "Find in Files"

drewnoakes:
Similar to other editors and IDEs, make the Ctrl+Shift+F keyboard shortcut show the "Find in Files" search panel, moving focus to the search text box and selecting any existing text to make it easier to type or paste a new search term in directly.


KirillOsenkov:
How did I not think of this

And I of all people have implemented Ctrl+Shift-F in another app

PR #910, opened 2025-12-03
Improve start page shadow in dark theme

drewnoakes:
In dark theme, the shadow was lighter than the background, so it was more of a glow than a shadow, which looked a bit off.

This approach uses opacity to blend with whatever background the theme chooses. I made it a bit smaller too, so it was more noticeable in dark mode.

Before

image

After

image


KirillOsenkov:
lol the glow was accidental but I kinda liked it :) ah well

KirillOsenkov (approved):

PR #916, opened 2026-01-18
Add redirecting button informing about LLM chat experimental release

JanKrivanek:
### What is this about?

image

Takes users to 

https://github.com/JanKrivanek/MSBuildStructuredLog/releases

### Why

Let's try to see if there is any interest in such feature at all


KirillOsenkov:
@jdom check out the latest version, it now redirects to VS Code

PR #928, opened 2026-02-08
Add themed Chrome titlebar

yuehuang010:
## Add a custom chromed title bar to match the dark/light theme.

Before:
image
image

After:
image
image


KirillOsenkov:
While I get the idea that it does look nicer with the custom title bar, the custom buttons look homemade and will be harder to support. 

For example, when you hover over the middle button, does it show the popup menu like the OS button does?

When the window is maximized and you move the mouse into the top right corner of the screen, and click, does it click on the close button?

What is the hover effect?

Basically there's a lot of custom behavior and support costs, and it just doesn't justify the added value.

It always hurts to decline PRs, but I'm already more than burned out maintaining this repo as is, so I hope for your understanding.

yuehuang010:
Understandable, I was a bit annoyed that there wasn't a WPF built-in support.

PR #955, opened 2026-07-09
Add MSBuild Server lifecycle node support (binlog format v27)

JanProvaznik:
Companion to [dotnet/msbuild#14156](https://github.com/dotnet/msbuild/pull/14156), which records how each build relates to the MSBuild Server node as a dedicated `MSBuildServerLifecycleEventArgs`, serialized under its own `BinaryLogRecordKind.MSBuildServerLifecycle` (binary-log **format version 27**).

This teaches the viewer to read and render that record.

### Changes
- **Read** the new `MSBuildServerLifecycle` record (`BuildEventArgsReader` + a vendored `MSBuildServerLifecycleEventArgs`).
- **Render** it as a dedicated `MSBuildServerNode` under a **"MSBuild Server"** folder, with a distinct icon (WPF `Generic.xaml` + Avalonia). The node exposes `Kind` (Spawned / Reused / NotUsed), `ShortLived`, `ProcessId`, `Reason` and `ReasonCode`.
- Also **read the `LoggersRegistered` record** (format v26). Without this, opening a v26/v27 binlog reports *"Skipped some data unknown to this version of Viewer"* for that record; it is now surfaced as an ordinary message.

### Behavior for the four server scenarios
| Scenario | Node |
|---|---|
| cold build | `Kind=Spawned` |
| next build | `Kind=Reused` (same PID) |
| `-nr:false` (non-`/mt`) | `Kind=NotUsed`, `ReasonCode=node-reuse-disabled` |
| `/mt -nr:false` | `Kind=Spawned`, `ShortLived=True` |

### Notes
- The record kind is **backward-compatible**: older viewers that predate it skip it gracefully (it is length-prefixed and its minimum-reader-version is unchanged). This PR simply lets the viewer render it instead of skipping.
- `MSBuildServerNode` serializes as a `Message` (like `TimedMessage`/`MessageWithLocation`) so `.buildlog` round-tripping is unaffected; the dedicated CLR type still drives the distinct icon.
- Draft pending the MSBuild side (dotnet/msbuild#14156) merging and the format version being finalized.

### DIFF
--- a/src/StructuredLogger/BinaryLogger/BinaryLogRecordKind.cs
+++ b/src/StructuredLogger/BinaryLogger/BinaryLogRecordKind.cs
@@ -37,5 +37,7 @@
         BuildCheckAcquisition,
         BuildSubmissionStarted,
         BuildCanceled,
+        LoggersRegistered,
+        MSBuildServerLifecycle,
     }
 }
--- a/src/StructuredLogViewer.Avalonia/App.xaml
+++ b/src/StructuredLogViewer.Avalonia/App.xaml
@@ -303,6 +303,19 @@
                            Opacity="{Binding IsLowRelevance, Converter={StaticResource RelevanceConverter}}" />
             </StackPanel>
         </DataTemplate>
+
+        <DataTemplate DataType="{x:Type l:MSBuildServerNode}">
+            <StackPanel Orientation="Horizontal">
+                <Rectangle Name="icon"
+                           Classes="nodeIcon"
+                           Stroke="{StaticResource MSBuildServerStroke}"
+                           Fill="{StaticResource MSBuildServerBrush}" />
+                <TextBlock Name="messageText"
+                           Classes="nodeText"
+                           Text="{Binding ShortenedText}"
+                           Opacity="{Binding IsLowRelevance, Converter={StaticResource RelevanceConverter}}" />
+            </StackPanel>
+        </DataTemplate>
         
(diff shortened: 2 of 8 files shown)


src/StructuredLogger/BinaryLogger/BinaryLogRecordKind.cs
```diff
@@ -37,5 +37,7 @@ public enum BinaryLogRecordKind
         BuildCheckAcquisition,
         BuildSubmissionStarted,
         BuildCanceled,
+        LoggersRegistered,
```
KirillOsenkov, line 40: what is LoggersRegistered? is it something I missed earlier?

KirillOsenkov:
if we confirm that we're absolutely sure we want to rev the binlog format, then this looks fine. If @baronfel says go ahead you can undraft and I can merge and publish the viewer first, then we can test binlogs produced by the MSBuild PR to make sure it works, then merge the MSBuild PR then.

JanProvaznik:
tested with local build of viewer 
image

src/StructuredLogger/BinaryLogger/BinaryLogRecordKind.cs
JanProvaznik: introduced in https://github.com/dotnet/msbuild/pull/13577

KirillOsenkov (approved):

PR #956, opened 2026-07-16
Recover truncated binlogs from killed or timed-out builds

nohwnd:
When a build that writes a `.binlog` is hard-killed or times out before it finalizes the log (test-host timeout killer, CI cancel that TerminateProcess-es, `taskkill /F`, OOM), the file is left with no `EndOfFile` marker and no gzip footer. Most of the build is already on disk, but reading it throws `EndOfStreamException` on the final partial record. The viewer surfaces that as a red "Error when reading the file" node with a raw stack trace, and the CLI paths (`BinlogTool`, raw `ReadRecords`) crash with `0xE0434352`. 

**This reads very harsh and as an internal error of the viewer, leading me to believe that the log is useless.** As a result a log that is mostly fine, looks corrupt and gets thrown away, which is exactly the log you want when a build times out.

This change softens it, and shows the user that this is an "expected" error state, that we know how to handle, and that we show them as much information as we can:

> Recovered a partial binlog. The build was stopped before the log was finalized, so it holds only the events collected up to that point — everything shown is valid.

The wording was chosen intentionally from multiple candidates to communicate that this expected error state (recovered) and that this is an error state (partial).

Because the fix is on the reader-side it it recovers logs from older MSBuild versions too, no writer change needed. Only the final partial record and whatever was still in the 32 KB buffer at kill time is lost.

Same `killed.binlog` (a real `dotnet build -bl` killed mid-build) before and after. 63,326 nodes recover either way, only the message changes.

Before:

before

After:

after

The raw exception and stack trace are gone, the yellow first-chance "Unable to read beyond the end of the stream" toast no longer fires because nothing throws, and `$error` finds one honest node. The build still shows as failed, which is correct, it was stopped.

**Command line**

`BinlogTool` reads the same records, so it hit the same truncation. Before this change it did not just show an error, it crashed the process:

```
binlogtool dumprecords --input:killed.binlog --include-rollup --include-total --exclude-details

RecordType, Count, Size[bytes]
==============================
Unhandled exception. System.IO.EndOfStreamException: Unable to read beyond the end of the stream.
   at System.IO.BinaryReader.ReadByte()
   at ...BuildEventArgsReader.ReadRaw(...) in BuildEventArgsReader.cs:line 185
   at ...BinLogReader.ChunkBinlogWithOffsets(...)+MoveNext() in BinLogReader.cs:line 333
   at BinlogTool.DumpRecords.RunSingle(...) in DumpRecords.cs:line 73
   at BinlogTool.Program.Main(...) in Program.cs:line 240
```

Exit code `-532462766` (`0xE0434352`, the CLR code for an unhandled managed exception).

After, the raw reader stops at a synthetic `EndOfFile` at the truncation point and the tool finishes normally:

```
binlogtool dumprecords --input:killed.binlog --include-rollup --include-total --exclude-details

...
ProjectFinished, 6, 27.00
TargetSkipped, 43, 66.95
EndOfFile, 1, 0.00

Total records: 10091
```

Exit code `0`.

I left the exit code at `0` on purpose. A finalized log of a build that failed to compile also returns `0` today, the build error is just an `Error` record in the stream:

```
binlogtool dumprecords --input:failed.binlog --include-rollup --include-total --exclude-details

Error, 1, 36.00
EndOfFile, 1, 0.00
Total records: 8797
```

Exit code `0`.

So on the command line the truncated log is now downgraded to the same level as any other build error. The tool no longer crashes, it reads what is there, and the exit code keeps the meaning it always had, that the tool processed the file and not that the build was error free. The one behavior that changed is that a truncated log used to abort with the `0xE0434352` crash code, and now returns `0` like any other readable log.

Verification: library, WPF viewer and BinlogTool all build clean. Added 11 regression tests in `BinaryLoggerTests.cs` covering the reader record path, the raw/CLI path, and the end-to-end gzip + header + `ReadBuild` seam (finalized logs stay clean, truncated logs at record-boundary and mid-record cuts recover a prefix and set the flag, and the user-facing error text is asserted). The 4 existing forward-compatibility reader tests still pass.

Related: dotnet/msbuild#3815. The writer-side counterpart (finalize on cancel, or periodic gzip member flush) would prevent the loss in the first place, this is the reader-side half that makes every already-broken log usable.

### DIFF
--- a/src/StructuredLogger.Tests/BinaryLoggerTests.cs
+++ b/src/StructuredLogger.Tests/BinaryLoggerTests.cs
@@ -438,5 +438,244 @@ namespace Microsoft.Build.UnitTests
             // There is nothing else in the stream
             memoryStream.Position.Should().Be(length);
         }
+
+        // ---------------------------------------------------------------------------------------
+        // Truncated / interrupted binlog recovery.
+        //
+        // A build that is hard-killed (test-run timeout killer, taskkill /F, CI job cancel, OOM,
+        // crash) never runs BinaryLogger.Shutdown(), so its .binlog is missing the trailing
+        // EndOfFile marker (and the gzip footer). The reader must recover every complete record
+        // written before the cut, stop cleanly instead of throwing, and flag the log as truncated
+        // so the viewer/CLI can tell the user the log was "recovered" and is "partial".
+        // ---------------------------------------------------------------------------------------
+
+        [Fact]
+        public void InterruptedLog_ReadStopsCleanlyAndFlagsTruncation()
+        {
+            // No EndOfFile marker => the producing build was interrupted before it finalized the log.
+            byte[] records = WriteEventRecords(count: 5, withEndOfFileMarker: false);
+
+            var (recovered, truncated, thrown) = ReadAllEvents(records);
+
+            thrown.Should().BeNull("a truncated log must stop cleanly, not throw");
+            recovered.Should().Be(5, "every complete record written before the cut is still recoverable");
+            truncated.Should().BeTrue("a stream that ends without an EndOfFile marker is a truncated log");
+        }
+
+        [Fact]
+        public void FinalizedLog_ReadDoesNotFlagTruncation()
+        {
+            // A cleanly shut-down build terminates the record stream with an EndOfFile marker.
+            byte[] records = WriteEventRecords(count: 5, withEndOfFileMarker: true);
+
+            var (recovered, truncated, thrown) = ReadAllEvents(records);
+
+            thrown.Should().BeNull();
+            recovered.Should().Be(5);
+            truncated.Should().BeFalse("an EndOfFile marker means the log was finalized - not truncated");
+        }
+
+        [Theory]
+        [InlineData(0.90)]
+        [InlineData(0.60)]
+        [InlineData(0.35)]
+        [InlineData(0.15)]
+        public void TruncatedLog_ReadRecoversPrefixWithoutThrowing(double keptFraction)
+        {
+            // Cut the stream at an arbitrary offset to hit both record-boundary and mid-record
+            // truncation. Whatever the cut, reading must never throw and must flag truncation.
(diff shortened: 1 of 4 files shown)


nohwnd:
For the record, here is what I tried and did not ship.

The first cut made truncation a yellow Warning node, with a `Build.Interrupted` tri-state and a "Recovered N records" count. It reads softer, but a build that got killed did fail, so dressing it up as a warning felt dishonest, and it meant adding behavior (a new build state) for what is really a messaging change. I went with a single error instead and reverted the `Interrupted` flag and the record counting.

tried a warning node

While chasing why a recovered log looks broken, there are actually two separate false alarms in the viewer, both from the same one truncation exception: the red "Error when reading the file" tree node (from `BinaryLog.ReadBuild`'s `OnException`), and a yellow first-chance toast "Unable to read beyond the end of the stream" (from `MainWindow`'s first-chance handler, which fires even though `Replay` catches the exception). Making `Read()` not throw removes both at once, the toast never fires because nothing is thrown, and the `OnException` node is replaced by the single "Recovered a partial binlog" node.

One nuance worth writing down: graceful Ctrl+C is fine already. When you Ctrl+C a `dotnet build -bl`, MSBuild raises `BuildCanceledEventArgs`, runs `EndBuild` and `BinaryLogger.Shutdown()`, so it writes `BuildFinished` and `EndOfFile` and the log is finalized even if you cancel mid-compile (verified twice). The truncated case is specifically the hard kill or timeout: test-host timeout killer, CI cancel that TerminateProcess-es, VS "Stop" force kill, OOM, `taskkill /F`. Those skip `Shutdown()`, so this fix is aimed at them.

I also have a working reconstruct prototype that splices the recovered records and re-appends `EndOfFile` and a fresh gzip footer, so external replay consumers (`msbuild --replay`, `dotnet build` re-read) that do not use this reader get a fully valid file. That is a separate `binlogtool repair` change, not in this PR.

KirillOsenkov (approved):

KirillOsenkov:
thank you!

PR #960, opened 2026-08-04
Include MSBuild node id with timing information

chwarr:
Sometimes it's useful to know which MSBuild node ran a given target/task. For example, to correlate with PID-based logging of msbuild.exe activity.

Include the node when "timing and duration" information is displayed, like from the "Show time and duration" context menu or the tooltip in the Timeline and Tracing views.

Here are some screenshots of where it shows up:

screenshot of the &quot;Show time and duration&quot; view
<br />
screenshot of the timeline view tooltip
<br />
screenshot of the tracing view tooltip

### DIFF
--- a/src/StructuredLogger/ObjectModel/TimedNode.cs
+++ b/src/StructuredLogger/ObjectModel/TimedNode.cs
@@ -52,7 +52,8 @@ namespace Microsoft.Build.Logging.StructuredLogger
 
             return $@"Start: {TextUtilities.Display(StartTime, displayDate: true, fullPrecision)}
 End: {TextUtilities.Display(EndTime, displayDate: true, fullPrecision)}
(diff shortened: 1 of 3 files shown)


src/StructuredLogger/ObjectModel/TimedNode.cs
chwarr, line 45: This no longer only gets time and duration text. Can/should this method be renamed to something like GetRuntimeInfoText?

KirillOsenkov: I'd say leave the method as it was, and have a new method that appends the node id, and have the caller call that instead.

KirillOsenkov:
I merged so it doesn't miss the next train. Thanks!

PR #961, opened 2026-08-11
Add filtered binary log replay

YuliiaKovalova:
## Summary

- add an opt-in `BinLogReader.EventFilter` callback with record kind and build event context metadata
- skip rejected event-specific payloads in version 18+ length-framed binlogs while continuing to process auxiliary records
- preserve required culture initialization for modern and legacy binlogs, and fall back to full deserialization for older formats and `TargetSkipped` records
- propagate filter callback exceptions while always completing the threaded replay worker
- cover compact reserialization, metadata, truncation, culture preambles, worker cleanup, and exception behavior with tests

## Motivation

Consumers that need only a subset of a large binary log currently deserialize every event. This API lets them reject records after reading lightweight common metadata, then connect accepted replay events to `BinaryLogger` to produce a compact, standalone filtered binlog.

## Validation

- 6 filtered-replay tests pass
- 98 remaining tests pass when excluding three environment-dependent baseline failures reproduced on unmodified `main`

### DIFF
--- a/src/StructuredLogger/BinaryLogger/BinLogReader.cs
+++ b/src/StructuredLogger/BinaryLogger/BinLogReader.cs
@@ -6,7 +6,6 @@ using System.Diagnostics;
 using System.IO;
 using System.IO.Compression;
 using System.Linq;
-using System.Runtime.ExceptionServices;
 using Microsoft.Build.Framework;
 
 namespace Microsoft.Build.Logging.StructuredLogger
@@ -113,9 +112,10 @@ namespace Microsoft.Build.Logging.StructuredLogger
                         {
                             instance = reader.Read(EventFilter);
                         }
-                        catch (BinaryLogEventFilterException ex)
+                        catch (BinaryLogEventFilterException)
                         {
-                            ExceptionDispatchInfo.Capture(ex.InnerException).Throw();
+                            // A failing filter callback is a caller bug, not a problem with the log:
+                            // abort the replay instead of reporting it via OnException and continuing.
                             throw;
                         }
                         catch (Exception ex)
@@ -176,9 +176,10 @@ namespace Microsoft.Build.Logging.StructuredLogger
                     {
                         instance = reader.Read(EventFilter);
                     }
-                    catch (BinaryLogEventFilterException ex)
+                    catch (BinaryLogEventFilterException)
                     {
-                        ExceptionDispatchInfo.Capture(ex.InnerException).Throw();
+                        // A failing filter callback is a caller bug, not a problem with the log:
+                        // abort the replay instead of reporting it via OnException and continuing.
                         throw;
                     }
                     catch (Exception ex)
--- a/src/StructuredLogger/BinaryLogger/BinLogReader.cs
+++ b/src/StructuredLogger/BinaryLogger/BinLogReader.cs
@@ -6,6 +6,7 @@ using System.Diagnostics;
 using System.IO;
 using System.IO.Compression;
 using System.Linq;
+using System.Runtime.ExceptionServices;
 using Microsoft.Build.Framework;
 
 namespace Microsoft.Build.Logging.StructuredLogger
@@ -26,6 +27,18 @@ namespace Microsoft.Build.Logging.StructuredLogger
         public event Action<IDictionary<string, string>, long> OnNameValueListRead;
         public event Action<int> OnFileFormatVersionRead;
         public event Action<IEnumerable<string>> OnStringDictionaryComplete;
+
+        /// <summary>
+        /// Gets or sets a filter that decides which build events are deserialized and dispatched.
+        /// </summary>
+        /// <remarks>
+        /// For length-framed binlogs, rejected events skip their type-specific payload without
+        /// deserializing it. Auxiliary records are always read so retained events can resolve their
+        /// string and name/value-list references. Reader preamble events required to interpret
+        /// subsequent events, such as the current UI culture marker, are retained regardless of the
+        /// filter. The filter is responsible for retaining a structurally consistent event set.
+        /// </remarks>
+        public BinaryLogEventFilter EventFilter { get; set; }
 
         /// <summary>
         /// Raised when there was an exception reading a record from the file.
@@ -96,7 +109,12 @@ namespace Microsoft.Build.Logging.StructuredLogger
 
                     try
                     {
-                        instance = reader.Read();
+                        instance = reader.Read(EventFilter);
+                    }
+                    catch (BinaryLogEventFilterException ex)
+                    {
+                        ExceptionDispatchInfo.Capture(ex.InnerException).Throw();
+                        throw;
                     }
                     catch (Exception ex)
                     {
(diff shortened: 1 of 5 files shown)


src/StructuredLogger/BinaryLogger/BuildEventArgsReader.cs
```diff
@@ -391,7 +462,17 @@ void HandleError(FormatErrorMessage msgFactory, bool noThrow, ReaderErrorType re
...
+            while (count > 0)
+            {
+                int read = _binaryReader.Read(buffer, 0, Math.Min(count, buffer.Length));
+                if (read == 0)
+                {
+                    throw new EndOfStreamException();
+                }
+
+                count -= read;
+            }
```
JanKrivanek, line 491: Why is this change needed?

src/StructuredLogger/BinaryLogger/BuildEventArgsReader.cs
```diff
@@ -275,9 +278,63 @@ private void CheckErrorsSubscribed()
...
+                        if (result != null && filterAfterDeserialization)
+                        {
+                            var metadata = new BinaryLogEventMetadata(
+                                recordKind,
+                                result.BuildEventContext,
+                                (result as TargetSkippedEventArgs)?.OriginalBuildEventContext);
+                            bool isRequiredPreamble =
+                                !sawCultureBeforeDeserialization && sawCulture;
+
+                            if (!isRequiredPreamble && !ApplyEventFilter(eventFilter!, metadata))
```
JanKrivanek: Is it sure that `eventFilter` cannot be null here?

src/StructuredLogger/BinaryLogger/BinLogReader.cs
```diff
@@ -90,39 +103,49 @@ public void Replay(Stream stream, Progress progress)
...
-                        progress.Report(new ProgressUpdate { Ratio = ratio, BufferLength = queue.Count });
+                        BuildEventArgs instance = null;
+
+                        try
+                        {
+                            instance = reader.Read(EventFilter);
+                        }
+                        catch (BinaryLogEventFilterException ex)
+                        {
+                            ExceptionDispatchInfo.Capture(ex.InnerException).Throw();
```
JanKrivanek: Why do we want to hide the wrapping exception?

JanKrivanek (approved):
Looks good and helpful. I added couple questions

If BuildEventArgsReader.cs is still in sync with dotnet/msbuild (and it should) - we should reflect the change there as well

src/StructuredLogger/BinaryLogger/BuildEventArgsReader.cs
YuliiaKovalova: `_binaryReader.BaseStream` is the `TransparentReadStream`, and its `Seek(..., SeekOrigin.Current)` funnels into `StreamExtensions.SkipBytes(throwOnEndOfStream: true)`, which reports a truncated stream as `InvalidDataException`. `Read()` can't tell that apart from a genuinely corrupt record, so it would go down the `HandleError`/`UnknownFormatOfEventData` path and a truncated log would fail hard instead of being recovered (`ReachedEndOfStreamPrematurely`). That matters now because the filter path skips whole records, so a truncated tail is much more likely to be hit inside `SkipBytes`.

Simplified in bdc2ba9e to reuse the shared helper with `throwOnEndOfStream: false` (plus a reused buffer) and translate the short read into `EndOfStreamException`, with a comment explaining why `Seek` isn't used. `Read_EventFilter_PreservesTruncationRecoveryInsideRejectedEvent` covers this.

src/StructuredLogger/BinaryLogger/BuildEventArgsReader.cs
YuliiaKovalova: It couldn't be null (`filterAfterDeserialization` was only ever `true` when `eventFilter != null`), but that was not obvious - hence the `!`. Reworked in bdc2ba9e so the two paths are held as non-null locals instead of a bool, which drops the null-forgiving operator entirely:

```csharp
BinaryLogEventFilter? filterBeforeDeserialization = null;
BinaryLogEventFilter? filterAfterDeserialization = null;
if (eventFilter != null)
{
    if (_fileFormatVersion < BinaryLogger.ForwardCompatibilityMinimalVersion ||
        recordKind == BinaryLogRecordKind.TargetSkipped)
    {
        filterAfterDeserialization = eventFilter;
    }
    else
    {
        filterBeforeDeserialization = eventFilter;
    }
}
```

src/StructuredLogger/BinaryLogger/BinLogReader.cs
YuliiaKovalova: Good point - no reason to hide it. The wrapper only exists to tell "the caller's filter threw" apart from "reading the log failed" (the latter is routed to `OnException` and the replay continues; the former must abort). It was being unwrapped just to keep the exception the consumer threw at the top of the chain, but that lost the context.

Fixed in bdc2ba9e: `BinaryLogEventFilterException` is now public and propagates as-is, with the filter's exception as `InnerException`. `BinLogReader` just rethrows it, so `ExceptionDispatchInfo` is gone. Test updated to assert on the wrapper + inner exception.

YuliiaKovalova:
@JanKrivanek re: keeping `BuildEventArgsReader.cs` in sync - the matching change is up as a draft at dotnet/msbuild#14693.

A few things differ there, driven by what msbuild's reader actually has:
- **No culture preamble special-casing.** `ProcessCultureMessage`/`Strings.Initialize` is viewer-only (`BuildEventArgsReader.Viewer.cs`), so the msbuild port drops that branch.
- **`SkipBytes` left untouched.** The change here exists to keep truncation recovery (`ReachedEndOfStreamPrematurely`) working when a rejected record is skipped; msbuild has no truncation recovery, so `Seek` stays as-is there.
- **Raw passthrough had to be handled.** `BinaryLogReplayEventSource` copies records verbatim when there are no structured subscribers, which would silently bypass the filter - so a set `EventFilter` now forces structured reading.
- **New public API** (`BinaryLogEventMetadata`, `BinaryLogEventFilter`, `BinaryLogEventFilterException`, the `Read` overload, `BinaryLogReplayEventSource.EventFilter`) plus a localized resource string, hence draft pending API review.

KirillOsenkov:
I'm guessing there's no perf overhead when no filter is specified? and we don't care much about perf when the filter is specified? would be good to see some numbers for large-ish binlogs.

KirillOsenkov:
Published in https://www.nuget.org/packages/MSBuild.StructuredLogger/2.3.244

YuliiaKovalova:
> Published in https://www.nuget.org/packages/MSBuild.StructuredLogger/2.3.244

I measured the dotnet/msbuild port (#14693) by replaying binlog: call Replay() => subscribe to AnyEventRaised  -> count events, and wrap the replay in  GC.GetTotalAllocatedBytes(precise: true).
 
There's a warmup pass first, and a forced GC before every iteration.
Scenarios are interleaved, and I reverse the order every other round so no run gets an advantage from sitting in a warm cache. 

1. No filter -> no overhead. 451.7 MB / 21.9 M events:
┌────────────────────────┬────────────┐
│                                    │     Events         │      Allocated bytes
├────────────────────────┼────────────┤
│ baseline                      │ 21,914,589      │  23,319,014,776     
│ with change, no filter │ 21,914,589      │  23,319,022,648      │
└────────────────────────┴────────────┘
Counts match exactly. Allocations differ by 7,872 B out of 23.3 GB (0.00003%).

2. With a filter (_allocations_, same log): 
- dropping ProjectEvaluation*  −36%; 
- dropping messages/diagnostics −24%; 
- keeping only start/finish records + errors + warnings −63%.

more info is here https://github.com/dotnet/msbuild/pull/14693

